PR #40, opened 2020-08-15
adder v1

Mikle25:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,21 @@
 
 function makeInfinityAdder() {
   // write code here
```
Andrii256, line 9: mindless comments are rubbish in your code

PR #43, opened 2020-08-16
solution

furude-rika:


src/makeInfinityAdder.js
```diff
@@ -4,9 +4,22 @@
  *
  * @return {function}
  */
+let sum = 0;
 
 function makeInfinityAdder() {
```
leonbohdan, line 9: give some parameter
you can write here like this - **makeInfinityAdder(x)**  with default value

src/makeInfinityAdder.js
```diff
@@ -4,9 +4,22 @@
...
-  // write code here
+  return (number) => {
+    if (number) {
+      sum += number;
+
+      return makeInfinityAdder();
```
leonbohdan, line 14: here we can write - **return makeInfinityAdder(x + number);** without sum += number

src/makeInfinityAdder.js
```diff
@@ -4,9 +4,22 @@
  *
  * @return {function}
  */
+let sum = 0;
 
 function makeInfinityAdder() {
-  // write code here
+  return (number) => {
```
leonbohdan, line 10: we can give some name for the inner function and return it after all

src/makeInfinityAdder.js
```diff
@@ -4,9 +4,22 @@
  *
  * @return {function}
  */
+let sum = 0;
```
leonbohdan, line 7: we don't need this variable because we can add our values to variable from parameter in **makeInfinityAdder(x)**

src/makeInfinityAdder.js
```diff
@@ -4,9 +4,22 @@
...
+    if (number) {
+      sum += number;
+
+      return makeInfinityAdder();
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
```
leonbohdan, line 20: we can return only value from parameter in **makeInfinityAdder**  without 
_const result = sum; 
sum = 0;_
because we have default value in **x** or value of numbers sum

furude-rika: thanks for all your advises! I'll fix my code using your recommendations!

PR #95, opened 2020-12-22
Infinite sum solution

VinnieJ-2k20:


Lara-l-art (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
...
+      totalSum = 0;
+
+      return result;
+    }
+
+    totalSum += number;
+
+    return function(nextNumber) {
+      return adder(nextNumber);
+    };
```
varseniuk, line 24: Why do you use this wrapper? All you need is to return the reference to your 'adder' function

varseniuk (changes requested):

src/makeInfinityAdder.js
VinnieJ-2k20: Yes, turns out it is much more simple solution. Thanks.

varseniuk (approved):

PR #96, opened 2020-12-22
Solution

vovador:


Serhii-Alieksieievych (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,28 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  const numbers = [];
```
vsm2k17, line 9: no need to save all numbers, just count how many times the function adder was called

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,28 @@
...
+
+        return lastInputOperator;
+      } else {
+        const accomulatedNumbers = [...numbers];
+
+        numbers.length = 0;
+
+        return accomulatedNumbers.reduce((acc, curV) => acc + curV, 0);
+      }
+    };
```
vsm2k17, line 29: it can be simplified

vsm2k17 (changes requested):

src/makeInfinityAdder.js
vsm2k17: sorry, not count(another task), sum, you need to store the sum

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+    if (number) {
+      numbers.push(number);
+
+      return InputOperator;
+    } else {
+      const accomulatedNumbers = [...numbers];
+
+      numbers.length = 0;
+
+      return accomulatedNumbers.reduce((acc, curV) => acc + curV, 0);
```
vsm2k17, line 21: instead of using reduce, you can store the sum of numbers as a number, not as an array (array takes much more machine memory than number)

vsm2k17 (changes requested):

vsm2k17 (approved):

PR #99, opened 2020-12-22
solution

troospieler:


AlexysUa (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let overAllsum = 0;
+  let counter = 0;
```
varseniuk, line 10: Why do you need this counter? All you need is your 'overAllSum', which you will return.
If there will be an argument, you add it to your overAllSum and return 'caller'. If there won't be any arguments (`if (x === undefined)`) you return the current overAllSum value and reset it to 0 and also return a 'caller'. 
So all you need - is only to check if x === undefined. No counter needed.
Try to simplify your code

varseniuk (changes requested):

varseniuk (approved):
All are good, but `returner` is  a very strange name)
Try not to use it in the future. Never)

PR #100, opened 2020-12-22
Solution

AlexysUa:


troospieler (approved):
good job!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,24 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result;
+  let sumOfNum = 0;
+
+  const add = (num) => {
+    if (num === undefined) {
+      result = sumOfNum;
```
vsm2k17, line 14: you can declare the result here, no need to save it in closures

vsm2k17 (approved):
good job

src/makeInfinityAdder.js
AlexysUa: Done!

PR #102, opened 2020-12-22
solution

xipholena:


sasha39612 (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const add = (x) => {
+    const currentSum = sum;
```
varseniuk, line 12: A better place for this variable 'currentSum' is inside your `else` statement. Because you actually don't need this variable if x !== undefined. So let's put it there not to declare it every time (even if there's no need for this).

varseniuk (changes requested):
All are good, but let's make it a little more efficient

varseniuk (approved):
Good 👍

PR #105, opened 2020-12-23
 solution

Pasha5555:


zhenya-mezhueva (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function ADD(num) {
```
```suggestion
  return function adder(num) {
```

Usually, we use uppercase for naming constants. Try not to name functions in uppercase

varseniuk (approved):
Good job.
Pay attention to my litttle remark

PR #107, opened 2020-12-23
solution

maryna-zinchenko:


alexmogwaiii (approved):
Круто!

varseniuk (changes requested):
Let's make your function much more simple.
You need to return a function that will check if there was the argument passed.
If it was - add it to the total sum and return the reference to the function itself.
If it wasn't - return the last value of the total sum, then annulate it (and, again, return a reference to itself, of course.

So that's all you need. No need to create two different functions inside and make check `if (arguments.length === 0)` two times.
Try to put all your logic inside one function that you will return.

varseniuk (approved):
Great 👍

PR #141, opened 2020-12-29
solution

NazarDovhanchuk:


viktor-morhun (approved):
gj!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+    }
+
+    const result = sum;
+
+    sum = 0;
+
+    return result;
+  };
+
+  return adder;
```
varseniuk, line 25:
```suggestion
  const adder = (num) => {
    if (num === undefined) {
    const result = sum;

    sum = 0;

    return result;
  };

    sum += num;

    return adder;
```

If you change your check to `if (num === undefined)` and swap your operations you can get rid of the duplicate `return adder` in your code

varseniuk (approved):
DRY. Take into consideration my comment

PR #211, opened 2021-05-03
Solution

nazarmatsevych:


IvanGrekov (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (number) {
```
VinnieJ-2k20, line 12: This  check will fail if we call `adder(0)`.
Use different condition.

VinnieJ-2k20 (changes requested):
Almost done.
Just fix one more thing.

VinnieJ-2k20 (approved):

PR #213, opened 2021-05-05
solution

Ihor-Karpyn:


igor-stasiv (approved):
Nice solution!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  const adder = (...n) => {
```
VinnieJ-2k20, line 10: Nice thinking with arguments, but here task says that each call of `adder` function will have either one or no arguments.
So, you don't need to use rest operator and simply check `if (n === undefined)`

VinnieJ-2k20 (changes requested):
Nice solution, but one detail may be improved.
Explanation in comment.

VinnieJ-2k20 (approved):
Good job 👍

PR #215, opened 2021-05-05
v_1.0.0

Andrey203:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(...args) {
+    if (args.length === 0) {
```
aliceblunt8, line 12: you can use single variable here, this function expects only one argument

```suggestion
  function adder(num) {
    if (!num) {
```

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #216, opened 2021-05-05
solution

Nazar-Kudatsky:


SkyPKWolf (approved):
Good job

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(num) {
+    if (!num) {
```
VinnieJ-2k20, line 12: If you use this check our program will not work correctly if we call `adder(0)`.

VinnieJ-2k20 (changes requested):
Almost finished.
One thing need to be fixed.

VinnieJ-2k20 (approved):

PR #217, opened 2021-05-05
Solution v0.1

SkyPKWolf:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let acc = 0;
+
+  const add = (num) => {
+    if (num) {
```
VinnieJ-2k20, line 12: if you call `adder(0)` your program will work incorrectly.
Let's change the condition check.

VinnieJ-2k20 (changes requested):
Good solution, only need to change one detail.

VinnieJ-2k20 (approved):

PR #219, opened 2021-05-06
solution

TimurHrynenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  const adder = (num) => {
+    if (num) {
```
Namart, line 11: will not work correct if num = 0

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+      sum += num;
+
+      return adder;
+    } else {
```
Namart, line 15: else is redundant

Namart (changes requested):

src/makeInfinityAdder.js
TimurHrynenko: yep, i checked and you were right)
i`ll fix this

src/makeInfinityAdder.js
TimurHrynenko: At first i didn`t get why, but now i understand

aliceblunt8 (approved):

PR #224, opened 2021-05-06
Solution

IvanVasylenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  const numbers = [];
+
+  const adder = (number) => {
+    if (number === undefined) {
+      const resultSum = numbers.reduce((sum, n) => sum + n, 0);
```
VinnieJ-2k20, line 13: You can use simple number as `sum` instead of array.
So you won't need to call `reduce` method each time.

VinnieJ-2k20 (changes requested):
Almost done. Fix just one thing.

aionova (approved):
Nice!

PR #225, opened 2021-05-06
infinite adder implemented

illisogori:


vitalya2626 (approved):
Everything looks great!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = function(number) {
+    const temp = sum;
```
VinnieJ-2k20, line 12: Not critical, but this operation would better be placed inside `if` block.

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+
+    if (number === undefined) {
+      sum = 0;
+
+      return temp;
+    }
+    sum += number;
```
VinnieJ-2k20, line 19:
```suggestion
    }
    
    sum += number;
```

VinnieJ-2k20 (approved):
Good job 👍 

Check the comments

PR #228, opened 2021-05-06
Solution

vova-myronenko:


illisogori (approved):
Good job.

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (number) {
```
illisogori, line 12:
```suggestion
    if (number === undefiend) {
```

VinnieJ-2k20: Agreed, it will not work correctly if you call `adder(0)`.

VinnieJ-2k20 (changes requested):
Almost done, need just one fix.

VinnieJ-2k20 (approved):
Good job 👍

PR #229, opened 2021-05-07
first solution

scarlettjohanssons:


progmari (approved):
🔥 👍

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,22 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  const add = (num) => {
+    if (!num) {
```
VinnieJ-2k20, line 13: This will work incorrectly if we call `add(0)`.
Need to use different check for `num` presence.

VinnieJ-2k20 (changes requested):
Good solution, it just needs one small fix.

VinnieJ-2k20 (approved):
👍

PR #231, opened 2021-05-07
Solution

bohdanklius:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
```
victor-arutiunov, line 9: Nice solution, I've the similar)

mykola-yarchuk (approved):
great

PR #236, opened 2021-05-09
Solution

MinchukO:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sumA = 0;
```
varseniuk, line 10:
```suggestion
  let sumAccumulator = 0;
```
Let's use more meaningful names. `sumAccumulator` or `finalSum` would be more semantic

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
```
varseniuk, line 9:
```suggestion
```

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
...
+  let sumA = 0;
+
+  const sum = (a) => {
+    if (a !== undefined) {
+      sumA += a;
+
+      return sum;
+    } else {
```
varseniuk, line 17:
```suggestion
    } 
```
`else` is redundant here. You are returning something in both cases

varseniuk (approved):
Good job!
Left some comments

PR #237, opened 2021-05-10
solution

taya294:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+
+      return oldSum;
+    } else {
+      sum += num;
+
+      return callback;
+    }
+  };
+
+  return callback;
```
varseniuk, line 24:
```suggestion
    }
      sum += num;

      return callback;
  };
```
You can get rid of `else` and it will mean that `sum += num` will execute every time when `num !== undefined` (`callback` will be returning as well)

varseniuk (approved):
Good job! 🚀 
Left one comment

PR #241, opened 2021-05-11
Solution

Grygoriy-Shytikov:


starzhitskiy (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (number) {
```
varseniuk, line 12:
```suggestion
    if (number !== undefined) {
```
What if I will pas the number `0` as an argument? '0' is boolean 'false', so this check will not pass and your function will return the final result. But what if it will be chained like `adder(5)(3)(0)(2)`? After your function will face `0` it will stop and will not add `2` after that because the `adder` function will not be returned. And instead of 5+3+0+2= 10, you will get an error. So you should check if the argument is present (`if (number !== undefined)`)

varseniuk (approved):
Nice work, but pay attention to my comment. It's important

PR #244, opened 2021-05-12
Solution

starzhitskiy:


Grygoriy-Shytikov (approved):
Looks Good!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const result = (num) => {
```
```suggestion
  return function result(num) => {
```
You can return the function directly. By the way, 'result' is the bad name for the function. You could name it like 'adder' or smth like this

varseniuk (approved):
Good 👍

PR #245, opened 2021-05-12
Solution

avgusta92:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
-}
+  let total = 0;
+
+  return function newFunc(addedValue) {
+    if (addedValue) {
```
varseniuk, line 12:
```suggestion
    if (addedValue !== undefined) {
```

What if I will pas the number '0' as an argument? '0' is boolean 'false', so this check will not pass and your function will return the final result. But what if it will be chained like `newFunc(5)(3)(0)(2)`? After your function will face '0' it will stop and will not add '2' after that because the adder function will not be returned. And instead of 5+3+0+2= 10, you will get an error. So you should check if the argument is present (`if (number !== undefined)`)
By the way - `newFunc` is a very bad name for the function. The name of the function should display what this function actually does.

varseniuk (approved):
Nice!
But left one important comment

PR #309, opened 2021-08-11
solution

VeronikaLushchik:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
...
+    if (a !== undefined) {
+      sum += a;
+
+      return getSum;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
```
ihor-jpeg, line 22: i think the else statement here is redundant

ihor-jpeg (approved):
Great code. I left one suggestion bellow;

herkoss (approved):
Great!💎✨

PR #316, opened 2021-08-14
solution

dima-tkachenko:


Anna-Tkachenko (approved):
👍🔥

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sumOfNumber = 0;
+  const adder = (number) => {
+    if (!number) {
```
VinnieJ-2k20, line 11: What if `number === 0` ???

VinnieJ-2k20 (changes requested):
Need to fix login in your code.

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sumOfNumber = 0;
```
danheim, line 9: name it easier
```suggestion
  let sum = 0;
```
```suggestion
  let total = 0;
```

danheim (changes requested):

VinnieJ-2k20 (approved):

PR #319, opened 2021-08-15
Recursive solution

lera-yaskevych:


Vova-Stoliar (approved):
Brilliantly💎

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (!number) {
```
VinnieJ-2k20, line 12: What if `number === 0` ???

VinnieJ-2k20 (changes requested):
Need to fix logic in your code.

src/makeInfinityAdder.js
lera-yaskevych: You're right. Fixed that!

VinnieJ-2k20 (approved):
Good job 👍

PR #321, opened 2021-08-16
Solution

sasha-morozov:


derelicttt (approved):
Good job!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (int) => {
+    if (!int) {
```
VinnieJ-2k20, line 12: What if `int === 0` ???

VinnieJ-2k20 (changes requested):
Need to fix logic in your solution.

VinnieJ-2k20 (approved):
Good job 👍

PR #323, opened 2021-08-16
solution

TarasHoliuk:


Andrii-Pinchuk (approved):
👍

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+
+  const adder = (x) => {
+    if (x !== undefined) {
+      amount += x;
+
+      return adder;
+    } else {
```
varseniuk, line 16:
```suggestion
    }
```
No need in this else here
Function will return something either way

varseniuk (approved):
🔥 
Pay attention to my comment

PR #325, opened 2021-08-16
Solution

Crystallord:


sasha-morozov (approved):
Nice solution!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let totalSum = 0;
+
+  const adder = (x) => {
```
```suggestion
  const adder = (num) => {
```
Better to use more meaningful names

varseniuk (approved):
🔥

PR #328, opened 2021-08-17
make adder

UlaDutchak:


sasha-morozov (approved):
Nice solution!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const add = (a) => {
```
```suggestion
  const add = (num) => {
```

Better to use more meaningful names

varseniuk (approved):
🔥

PR #334, opened 2021-08-19
solution by set1one

set1one:


MaksymBeliever (approved):
Everything's ok 👍

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
...
+      sum += num;
+    }
+
+    if (num === undefined) {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
varseniuk, line 22:
```suggestion
    if (num === undefined) {
      const result = sum;

      sum = 0;

      return result;
    }
    
    sum += num;
```
Let's swap these statements by places and get rid of redundant `if`

varseniuk (changes requested):

PR #374, opened 2021-10-19
Solution

MariiaObraztsova:


TRSKZK (approved):
Well done!!!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let totalSum = 0;
+
+  return function adder(number) {
+    if (!number) {
```
VinnieJ-2k20, line 12: It won't work with 0 as argument. Choose another check.

PR #375, opened 2021-10-19
solved

arielg0rl:


DionStrzh (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,18 @@
...
+  return function adder(x) {
+    if (x) {
+      num += x;
+      return adder;
+    }
+    if (!x) {
+      let result = num;
+      num = 0;
+      return result;
+    }
```
VinnieJ-2k20, line 19: This `if` can be removed. If code reaches this part, it will always be executed.

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,19 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let num = 0;
+
+  return function adder(x) {
+    if (x) {
+      num += x;
+
+      return adder;
+    }
```
VinnieJ-2k20, line 16: This won't work if we pass 0 as argument.
Choose another check.

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let num = 0;
+
+  return function adder(x) {
+    if (x || x === 0) {
```
VinnieJ-2k20, line 12: This is better.
```suggestion
    if (x !== undefined) {
```

PR #381, opened 2021-10-20
solution

HaqpaHR:


DarinaSavytska (approved):

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,21 @@
 
 function makeInfinityAdder() {
   // write code here
+  let result = 0;
+
+  return function sum(num) {
+    if (!num) {
```
VinnieJ-2k20, line 13: This won't work correctly if we pass 0 as argument.
Choose another check.

VinnieJ-2k20 (changes requested):

PR #382, opened 2021-10-20
[JS] Infinite-Sum-Closures STRZH

DionStrzh:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let currentSum = 0;
+
+  return function adder(number) {
+    if (!number) {
```
VinnieJ-2k20, line 12: This condition is incorrect, it will return `true` if we pass 0 to function.

PR #398, opened 2021-11-15
Solution

dariia-romanova:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+
+  const adder = (num) => {
+    if (num || num === 0) {
+      amount += num;
+
+      return adder;
+    }
```
dmtrhrytsak, line 16: If I'm right, you check here whether the argument `num` was passed?

```
 if (num || num === 0) {
      amount += num;

      return adder;
    }
```

And you realized that if a user passes 0 it won't add this num, but it should) So you added another checking for that `num === 0`. But, wouldn't it be more obvious if you were checking the number of passed arguments themselves? You could do that through `arguments.length` 

One important note here is that arrow functions don't have `arguments` property, so we should use functions instead)

dmtrhrytsak (approved):
This script is so durable that I'm pretty sure It could add numbers non-stop for the rest of my days) Awesome!

src/makeInfinityAdder.js
dariia-romanova: Yeah, I was wondering why arguments property doesn't work.
arguments.length is obviously better choice, thanks)

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+
+  const adder = function(num) {
+    if (arguments.length !== 0) {
```
ihor-jpeg, line 12: this is correct. an alternative would be `num !== undefined`

ihor-jpeg (approved):
Good work!

ihor-jpeg (approved):

ihor-jpeg (changes requested):
.

ihor-jpeg (approved):

PR #400, opened 2021-11-17
solution was added

JamesSunderlanddd:


SerhiiKirik (approved):
Well done 🔥

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(number) {
+    if (arguments[0] !== undefined) {
```
SerhiiKirik, line 12: bytheway, you can use an alternative it would be
```suggestion
    if (number !== undefined) {
```

JamesSunderlanddd: @SerhiiKirik I agree, it looks better this way and also easier to understand, thank you!)

SerhiiKirik: @JamesSunderlanddd your welcome 😉

aionova: agree

aionova (approved):

PR #403, opened 2021-11-18
infinityAdder

Vitaliy-0:


mariia-bila (approved):
LGTM 👍

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
...
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  let count = 0;
+
+  const sumarise = (a) => {
+    if (count > 0) {
+      sum = 0;
+    }
```
ihor-jpeg, line 15: i'm trying to understand what this part does

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  let count = 0;
```
ihor-jpeg, line 10: what does count do here?

ihor-jpeg (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,27 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const add = (a) => {
```
AlinaMatuschak, line 11: you should return a function that get a number. If a number exists the function updates your sum value and returns the add function. If a number doesn’t exist you reset the sum value and return the sum value before reset

for this task, you don't need a return function that returns another new function

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):
🤘

PR #410, opened 2021-11-20
First solution

WorkAccCom:


msaikun (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
-  // write code here
+  let numbers = [];
+
+  const adder = (number) => {
+    if (number !== undefined) {
+      numbers.push(number);
+
+      return adder;
+    } else {
+      const result = numbers.reduce((sum, num) => sum + num, 0);
```
AlinaMatuschak, line 17: Nice variant. But there you make unnecessary iterations. You can create number variable instead numbers and update a value

AlinaMatuschak (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (number !== undefined) {
+      sum += number;
+
+      return adder;
+    } else {
+      const result = sum;
```
AlinaMatuschak, line 17: Is it necessary to use else statement if the body of the previous condition returns something?
```suggestion
      return adder;
    }
    
    const result = sum;
```

AlinaMatuschak (approved):

PR #414, opened 2021-11-21
Solution

volodymyr-soltys97:


lera-permiakova (approved):
Perfect!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result = 0;
+
+  return function adder(number1) {
```
AlinaMatuschak, line 11: you should create a function that get a number. If a number exists the function updates your sum value and returns the adder function. If a number doesn’t exist you reset the sum value and return the sum value before reset

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):
🚀

PR #415, opened 2021-11-21
Solution

lera-permiakova:


volodymyr-soltys97 (approved):
Great job!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (a) => {
+    if (a !== undefined) {
```
AlinaMatuschak, line 12:
```suggestion
    if (a) {
```

AlinaMatuschak (approved):
Great 👍

src/makeInfinityAdder.js
lera-permiakova: But if I'll do this, than with '0' input won't work correctly

AlinaMatuschak: || a === 0

lera-permiakova: Then why this better than a !== undefined? :)

PR #417, opened 2021-11-22
Solution of task-inifinite-sum-closures

Dima-Ziakhor:


Nadiyahr (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (x) => {
+    if (!isFinite(x)) {
```
AlinaMatuschak, line 12: Nice variant. But also you can use equations
```suggestion
    if (!x && x !== 0) {
```

AlinaMatuschak (approved):
🔥

PR #422, opened 2021-11-22
Solution

OlhaSpring:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+    if (value === undefined) {
+      const result = cash;
+
+      cash = 0;
+
+      return result;
+    }
+    cash += value;
+
+    return (...args) => adder(args[0]);
```
ArtemTopolskyi, line 21: do you really need rest operator here?

ArtemTopolskyi (approved):

PR #423, opened 2021-11-22
solution: inifinite-sum-closures

lanebx:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
-  // write code here
+  let currentSum = 0;
+  let copyCurrentSum;
+
+  function foo(value) {
+    if (value !== undefined) {
+      currentSum += value;
+
+      return foo;
+    } else {
```
IrynaHerasymchuk, line 17: No need for 'else' if you return something in your if condition 😉

IrynaHerasymchuk (approved):
Good job!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let currentSum = 0;
+  let copyCurrentSum;
```
AlinaMatuschak, line 10: you use this variable just at one block, so you can create there constant

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+
+  function foo(value) {
+    if (value !== undefined) {
+      currentSum += value;
+
+      return foo;
+    } else {
+      copyCurrentSum = currentSum;
```
AlinaMatuschak, line 18:
```suggestion
      const copyCurrentSum = currentSum;
```

AlinaMatuschak (changes requested):
fix @IrynaHerasymchuk comment

AlinaMatuschak (approved):
🔥

PR #426, opened 2021-11-23
solution

sviatoslav-myhal:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (i) => {
+    if (!i && i !== 0) {
```
SerhiiKirik, line 12: You can just 
```suggestion
    if (i !== undefined) {
```

SerhiiKirik (approved):
Nice 🔥

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (i) => {
```
illiarerun, line 11: What is `i`? That's not some type of index that's number that you gonna add

illiarerun (changes requested):
Good job 🔥  But make sure to fix this one:

illiarerun (approved):
Good job 🔥

PR #443, opened 2021-12-23
 done

Niki-veb-dev:


Basil-Panasiuk (approved):

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
...
+    if (value === undefined) {
+      const sumResult = sum;
+
+      sum = 0;
+
+      return sumResult;
+    } else {
+      sum += value;
+
+      return add;
```
Olexandr6, line 22: Avoid `else` statement, you can simplify it without `else`

Olexandr6 (changes requested):
In general, your solution is good 👍  Check comments 👇

PR #447, opened 2021-12-24
Solution

Bohdan-Maksymiuk:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,24 @@
...
+  const adder = (x) => {
+    let sumInner = sumOuter;
+
+    if (x === undefined) {
+      sumOuter = 0;
+
+      return sumInner;
+    } else {
+      sumInner += x;
+      sumOuter = sumInner;
```
HrynykNataliia, line 20: можна скоротити до sumOuter += x;

HrynykNataliia (approved):

src/makeInfinityAdder.js
Olexandr6: Agree

Olexandr6 (changes requested):
Check comments 👇

Olexandr6 (approved):

PR #453, opened 2021-12-27
solution

vladyslav-harkusha:


sergeyglazgo (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+  return function adder(value) {
+    if (typeof value === 'number') {
+      buffer += value;
+    } else {
+      const sum = buffer;
+
+      buffer = 0;
+
+      return sum;
+    }
```

G-YaSha (changes requested):
In general, your solution is good 👍 Check comments 👇

G-YaSha (approved):

PR #454, opened 2021-12-27
jkrelhgfrhfgkn;xehnghftdhg

destroythemoon:


sergeyglazgo (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+  return function adder(num) {
+    if (num === undefined) {
+      const techSum = sum;
+
+      sum = 0;
+
+      return techSum;
+    } else {
+      sum += num;
+    }
```

Olexandr6 (changes requested):
In general, your solution is good 👍 Check comments 👇

PR #484, opened 2022-02-01
Complite

NovikovM18:


Tsekhmister (approved):
👍

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,22 @@
...
+
+  function adder(...args) {
+    if (args.length < 1) {
+      const result = amount;
+
+      amount = 0;
+
+      return result;
+    }
+    amount += args.reduce((sum, number) => sum + number, 0);
```
Anna-Tkachenko, line 20: Good job
In examples this function get only one number, so reduce is redundant here, you can just add one value to amount

Anna-Tkachenko (approved):

PR #487, opened 2022-02-04
Solution-1

MelnykA:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  function adder(x) {
```
aliceblunt8, line 9: only one function should be returned from the makeInfinityAdder

aliceblunt8 (changes requested):
only one function should be returned from the

danylo-onishchuk (approved):

PR #494, opened 2022-02-16
Multiple function calls

aspurgilus:


SanyaBratashchuk (approved):
interesting solution, but you can solve it using 1 adder function. Maybe it would be simplier solution

aspurgilus:
> interesting solution, but you can solve it using 1 adder function. Maybe it would be simplier solution

Yes, I changed my code. It much better now.

solaryasha (approved):

PR #518, opened 2022-03-21
Finish hw.

dant0n:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  const sum = (x) => {
+    if (typeof x !== 'undefined') {
```
ArtemTopolskyi, line 10: you don't need 'typeof' here
```suggestion
    if (x !== undefined) {
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      return (y) => {
+        if (typeof y !== 'undefined') {
+          return sum(x + y);
+        } else {
+          return x;
+        }
+      };
+    }
+
+    return 0;
```
ArtemTopolskyi, line 20: I think it's better to swap this statements and return 0 if (x === undefined), else - do main logic

ArtemTopolskyi (changes requested):

src/makeInfinityAdder.js
dant0n: makes sense to fail fast, thanks

ArtemTopolskyi (approved):
:fire:

PR #530, opened 2022-04-06
Solution

yaremenko-maksym:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,29 @@
...
+    if (x === undefined) {
+      sum = 0;
+
+      return sum;
+    }
+
+    sum += x;
+
+    return returnFunction;
+  };
```
AlinaMatuschak, line 31: this functions really similar. Can you create just one function?

AlinaMatuschak (approved):

AlinaMatuschak (changes requested):
try implement one function 👆

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
+      sum += x;
+
+      return returnFunction;
+    }
```
AlinaMatuschak, line 22: if x doesn't pass, the first condition converts to true and you return result. Everything that after return won’t run

```suggestion
      return result;
    }
    
    sum += x;

    return returnFunction;
```

AlinaMatuschak (approved):
Nice work 🚀

PR #550, opened 2022-04-16
Solution

YegorKochetkov:


Yurii837 (approved):

src/makeInfinityAdder.js
```diff
@@ -5,8 +5,26 @@
...
+
+      return result;
+    }
+
+    sum += num2;
+
+    return foo;
+  };
+
+  foo.valueOf = () => sum;
```
Yurii837, line 25: You were used valueOf, offbeat solution. Nace.

Yurii837 (approved):

src/makeInfinityAdder.js
YegorKochetkov: it's good that you reminded me, it's not needed here )

MaksimusCRS (approved):

PR #557, opened 2022-04-22
solution

vince15ua:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  let temp = 0;
```
aliceblunt8, line 10: why do you need temp? just use sum

aliceblunt8 (changes requested):

src/makeInfinityAdder.js
vince15ua: I cant understand how at the same time i must return sum and set sum = 0 to clear amount...

ArturErmolenko (approved):
👍

PR #564, opened 2022-05-08
solution

vlad-demchuk:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,29 @@
...
+      const clear = sum;
+
+      sum = 0;
+
+      return clear;
+    }
+
+    sum += current;
+
+    return function(input) {
```
alexandra-protyanova, line 22: I don't think you need second function

alexandra-protyanova (changes requested):

Denchepornyuk (approved):

PR #565, opened 2022-05-08
Implemented solution

dellachaise:


alexandra-protyanova (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  const add = (number) => {
+    if (number !== undefined) {
+      return (number2) => {
+        if (number2 !== undefined) {
+          return add(number + number2);
```
Denchepornyuk, line 13: create a variable above `add` function and use it inside add function for adding `number` and if `number` is undefined return the calculated value

the second function you can remove

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #572, opened 2022-05-11
infinite sum solution

koshachina:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,25 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  const infsum = (x) => {
```
ArtemTopolskyi, line 12:
```suggestion
  const infiniteAdder = (x) => {
```

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,25 @@
...
+    if (x || x === 0) {
```
ArtemTopolskyi, line 13:
```suggestion
    if (x !== undefined) {
```

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,25 @@
...
+      sum += x;
+
+      return function(y) {
+        return infsum(y);
+      };
```
ArtemTopolskyi, line 18: return infiniteAdder;

ArtemTopolskyi (changes requested):

PR #574, opened 2022-05-11
It works!!!

valerii-derkach:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  const adderFn = (value) => {
+    if (value === undefined) {
```
SavMary, line 10: you have done a double check the undefined. I think, you can do it without function adderFn

SavMary (approved):

Denchepornyuk (approved):

PR #575, opened 2022-05-11
Solution

amitpatiljc:


andreisucman (approved):
Good job. Again. 👍 

Maybe there's no need to create the finalSum in the outermost functions as it needs to exist only if the number === undefined. Though this isn't a mistake.

meforoma (approved):

PR #582, opened 2022-05-12
make Infinity Adder

kitaminary:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,24 @@
...
+  let counter = 0;
+
+  const adder = (num) => {
+    if (num === undefined && counter === 0) {
+      counter++;
+
+      return sum;
+    } else if (counter > 0) {
+      return 0;
+    }
```
Anna-Tkachenko, line 20: counter variable is redundant here
you shouldn't store the num of function calls
if num = undefined - you should return sum and reset this value
if num != undefined - you should add num to the sum and return the function

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #584, opened 2022-05-12
infinite sum

elkhayate:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,29 @@
...
+  let sum = 0;
+
+  const adder = (num) => {
+    const newSum = sum;
+
+    if (num !== undefined) {
+      sum += num;
+
+      return adder;
+    }
```
59MrRobot, line 18: This part is redundant in my opinion so I would suggest checking the opposite condition with the if statement below

59MrRobot (changes requested):

src/makeInfinityAdder.js
VinnieJ-2k20: Agree, logic of your code can be simplified.
Too many repeated code,

VinnieJ-2k20 (changes requested):

59MrRobot (approved):
Good job!

ihor-jpeg (approved):

PR #586, opened 2022-05-13
Solution

PRO-GRAM-MER:


59MrRobot (changes requested):

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let finalResult = 0;
+
+  const adder = (number) => {
+    if (!number && number !== 0) {
```
VinnieJ-2k20, line 13: There is more simple option for this check.
```suggestion
    if (number === undefined) {
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #588, opened 2022-05-16
solution

Uktam19980416:


59MrRobot (approved):
Good job

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let total = 0;
+
+  const adder = (...args) => {
```
VinnieJ-2k20, line 12: There are 0 or 1 arguments there. You don't need to use args with rest.

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
...
+    if (args.length === 0) {
+      let copySum = total;
+
+      total = 0;
+
+      return copySum;
+    }
+
+    total += +args;
```
VinnieJ-2k20, line 21: And you can avoid this.

VinnieJ-2k20 (changes requested):
Your code works, but it can be improved.

VinnieJ-2k20 (approved):
Good job :

PR #601, opened 2022-05-30
Solution v1.0

allesdmytro:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(x = '') {
+    if (x.length === 0) {
```
Bohdan-Maksymiuk, line 12:
```suggestion
  return function adder(x = '') {
    if (x.length === 0) {
```
why default `''` and check for `x.length`

Bohdan-Maksymiuk (changes requested):
we are working with numbers here

allesdmytro:
Fixed crutch with variable condition check

Bohdan-Maksymiuk (approved):

PR #613, opened 2022-06-12
Infinite sum closures

vsupruniuk:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,29 @@
...
+        sum = 0;
+
+        return result;
+      }
+
+      sum += newNum;
+
+      return newAdder;
+    };
+  };
```
Olexandr6, line 31: Actually, you don't need 2 nested functions in this task. Let's try to implement with only one function.

Olexandr6 (changes requested):
Let's improve your solution

PR #621, opened 2022-06-14
solution

vinnibea:
Still a little bit tough to understand it fully, but it must work...

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -7,6 +7,27 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+  let count = 0;
+
+  const add = (...args) => {
+    for (const x of args) {
+      sum = sum + x;
+      count++;
+
+      return add;
+    }
+
+    if (count === 0) {
+      sum = 0;
+    }
+
+    count = 0;
+
+    return sum;
+  };
+
+  return add;
 }
 
 module.exports = makeInfinityAdder;
--- a/package-lock.json
+++ b/package-lock.json
@@ -463,13 +463,248 @@
       }
     },
     "@mate-academy/scripts": {
-      "version": "0.3.5",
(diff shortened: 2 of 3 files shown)


vinnibea:
```
function adder() {
 let sum = 0;
 let count = 1;
  add = (...arguments) => {

      if (arguments.length > 1){
          for (const x of arguments) {
           sum += x;
       }

       return add;
      }

      for (const x of arguments) {
           sum += x;
          count--;
          return add;
      }
      

      if(count === 1) {
          sum = 0;
      }

      count = 0;

      return sum;
   }

    return add;
   }
```

and in case if you want to add multiple amount of numbers in one call

vinnibea:
```
function adder() {
 let sum = 0;
 let count = 0;
  add = (...arguments) => {
      
      

      if (arguments.length > 1){
          for (const x of arguments) {
           sum += x;
              count++;
       }

       return add;
      }

      for (const x of arguments) {
          sum += x;
          count++;
          return add;
      }

      if (count === 0) {
              sum = 0;
              return sum;
          }

      count = 0;

      return sum;
   }

    return add;
   }
```

just in case if you want to add multiple numbers in one call like this `add(100, 100) (100, 100)() === 400`

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,27 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+  let count = 0;
+
+  const add = (...args) => {
+    for (const x of args) {
```
stanislav-topikha, line 14: No need for a `for loop`, just `if` will do the job

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,27 @@
...
+      sum = sum + x;
```
stanislav-topikha, line 15: `sum += x`

stanislav-topikha (changes requested):
Looks great 🔥 
But check the comments ⬇️

mykola-koval (approved):
👍

PR #622, opened 2022-06-15
add solution

Galyopa:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (x) => {
```
Angelion541, line 11: maybe is a good idea to rename it something like => value

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+    if (x === undefined) {
```
mykola-koval, line 12:
```suggestion
    if (!x) {
```

or maybe even use && operator for this

mykola-koval (approved):

PR #628, opened 2022-06-16
Solution

Tania-Troshchuk:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      return 0;
+    }
+
+    return b => {
+      if (b === undefined) {
+        return a;
+      }
+
+      return adder(a + b);
+    };
```
aliceblunt8, line 20: no need to create another function, only one function (adder) is allowed

create a variable sum inside makeInfinityAdder and add to it each number passed in the arguments until it encounters a call without parameters

aliceblunt8 (changes requested):

vkryvytskyy (approved):

PR #634, opened 2022-06-19
Solution

ghost:


Angelion541 (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (number === undefined) {
+      const tmp = sum;
```
illiarerun, line 13: let's work on naming a bit. it is unclear what `tmp` means, try to use something more meaningful and without overshortening

illiarerun (changes requested):
GJ, one fix left:

jstmpelowycz (approved):

PR #636, opened 2022-06-19
Adder

AndrewKhorsun:


Angelion541 (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      const totalAmount = sum;
+
+      sum = 0;
+
+      return totalAmount;
+    } else {
+      sum += number;
+
+      return adder;
+    }
```
illiarerun, line 22: you don't need `else` here, anyways you return value from previous `if`
```suggestion
    }
    
    sum += number;

    return adder;
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+    if (number === undefined) {
+      const totalAmount = sum;
+
+      sum = 0;
+
+      return totalAmount;
+    }
+      sum += number;
+
+      return adder;
```
jstmpelowycz, line 21:
```suggestion
    if (number === undefined) {
      const totalAmount = sum;

      sum = 0;

      return totalAmount;
    }
    
    sum += number;

    return adder;
```

jstmpelowycz (approved):

PR #637, opened 2022-06-19
solved

Grodh06:


Angelion541 (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result = 0;
+
+  const nextValue = function(n) {
+    const amount = result;
```
illiarerun, line 12: you only need it if there is no `n`, so move it to if block, don't create unnecessary variables

illiarerun (changes requested):
GJ, one little fix left:

jstmpelowycz (approved):

PR #639, opened 2022-06-19
solution

illavodonis:


yanaost (approved):
looks good) except that variable 'res' could be a 'result' or smth more informative)

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  const addCalculate = (number) => {
```
illiarerun, line 10:
```suggestion
  let sum = 0;
  
  const addCalculate = (number) => {
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+    if (typeof number !== 'number') {
+      const res = sum;
```
illiarerun, line 12: `res` is somewhat okay, but I'd change it to something more meaningful

illiarerun (approved):
GJ, left some comments:

PR #643, opened 2022-06-20
Solution

MariiaNH:
Solution created

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let temporarySum = 0;
+
+  return function add(input) {
+    if (typeof input === 'undefined') {
+      const temporarySumCopy = temporarySum;
+
+      temporarySum = 0;
+
+      return temporarySumCopy;
+    } else {
+      temporarySum += input;
+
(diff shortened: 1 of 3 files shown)


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let temporarySum = 0;
+
+  return function add(input) {
+    if (typeof input === 'undefined') {
```
illiarerun, line 12: would be easier to do it like:
```suggestion
    if (input === undefined) {
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      const temporarySumCopy = temporarySum;
+
+      temporarySum = 0;
+
+      return temporarySumCopy;
+    } else {
+      temporarySum += input;
+
+      return add;
+    }
```
illiarerun, line 22: you don't need else here, anyways you return value from previous if
```suggestion
    }
   
    temporarySum += input;

    return add;
```

illiarerun (changes requested):
GJ, now let's work on fixes:

jstmpelowycz (approved):

PR #645, opened 2022-06-20
Add inifinite sum

ghost:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+      const result = count;
+
+      count = 0;
+
+      return result;
+    } else {
+      count += numbers;
+
+      return adder;
+    }
```
illiarerun, line 21: you don't need else here, anyways you return value from previous if

```suggestion
    }
   
    count += numbers;

    return adder;
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let count = 0;
```
illiarerun, line 9: that's not a `count`, it's a sum

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+  const adder = (numbers) => {
```
illiarerun, line 10: it's only 1 number, why you're using plural name here?
```suggestion
  let count = 0;
  
  const adder = (number) => {
```

illiarerun (changes requested):
GJ, let's work on fixes now:

jstmpelowycz (approved):

PR #646, opened 2022-06-21
first version

DmitriiPirohov:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const add = (args) => {
```
jstmpelowycz, line 11:
```suggestion
  const add = (arg) => {
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
...
+    let resultSum = 0;
+
+    if (typeof (args) !== 'number') {
```
jstmpelowycz, line 14: suggestion: I'd rather do it this way
```suggestion
    if (typeof args === 'undefined') {
```

jstmpelowycz (approved):

PR #665, opened 2022-07-07
Solution

vyurkov:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result = 0;
+
+  const add = (value) => {
+    if (!value && value !== 0) {
```
valerii-derkach, line 12: double check?
```suggestion
    if (value === undefined) {
```

valerii-derkach (changes requested):

src/makeInfinityAdder.js
vyurkov: My mistake. I was thinking about checking the passed parameter, which should be checked after this. Something like this:

PR #677, opened 2022-07-13
Solution

Kras4iy:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,27 @@
...
+  let sum = 0;
+  let emptyArgCall = false;
+
+  const addition = function(number) {
+    if (arguments.length === 0) {
+      if (!emptyArgCall) {
+        emptyArgCall = true;
+
+        return sum;
+      }
```
oleksandr-kotliarov, line 18:
```suggestion
  const addition = function(number) {
    if (arguments.length === 0) {
      const result = sum;

      sum = 0;

      return result;
    }
```

also works and looks simpler but it's up to you

oleksandr-kotliarov (approved):

src/makeInfinityAdder.js
Ihor-Karpyn: Agree let's use this advice

Ihor-Karpyn (changes requested):

Ihor-Karpyn (approved):

PR #679, opened 2022-07-15
Develop

denenergi:


src/makeInfinityAdder.js
```diff
@@ -5,8 +5,28 @@
...
+      return next;
+    }
+
+    if (!numberSecond) {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
VETALEK, line 24: you can combine theses 2 `if`s in one `if else`

VETALEK (changes requested):
Good start 💪

src/makeInfinityAdder.js
```diff
@@ -5,8 +5,26 @@
...
-  // write code here
+function makeInfinityAdder(numberFirst) {
+  let sum = (!numberFirst) ? 0 : numberFirst;
+
+  function next(numberSecond) {
+    if (numberSecond !== undefined) {
+      sum += numberSecond;
+
+      return next;
+    } else if (!numberSecond) {
```
VETALEK, line 16:
```suggestion
    } else {
```

VETALEK (changes requested):
Almost 💪

VETALEK (approved):
🚀

PR #683, opened 2022-07-16
Solution by DK

DimaK123:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const func = (num) => {
```
valerii-derkach, line 11: name variables and functions semantically, what they mean or do, not just `func`, it is not good practice. e.g. `adder`

valerii-derkach (changes requested):

src/makeInfinityAdder.js
DimaK123: Done

ArturErmolenko (approved):
Well done 👍

PR #693, opened 2022-07-18
Solution

yaroslav-lizogub:


roman-romanenko (approved):
Good 👍🏻 but hard to understand your solution for me 🤯

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,15 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  const adder = (sum) => (
+    sum ? (num) => (
+      num !== undefined
+        ? adder(sum + num)
```
solaryasha, line 12: Avoid using nested ternary operators. It's harder to read your code

solaryasha (changes requested):

PR #704, opened 2022-07-25
task

yura-atamaniuk:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
...
+        }
+      };
+
+      return arr;
+    }
+
+    return 0;
+  };
+
+  return addRol;
```
Bohdan-Maksymiuk, line 26: rename variables and use clear naming for them
and get rid of arrow function - you can store your sum in `makeInfinityAdder` directly

Bohdan-Maksymiuk (changes requested):

alexandra-protyanova (approved):

PR #713, opened 2022-08-03
Infinity Adder task solution

juliiaap:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,35 @@
...
+      } else {
+        sum += y;
+
+        return addition;
+      }
+    };
+  };
+
+  return addition;
 }
```
valerii-derkach, line 39: no needed else statement after return
```suggestion
  const addition = (x) => {
    if (x === undefined) {
      const result = sum;

      sum = 0;

      return result;
    } 
      sum += x;
    
        return addition;
    };
  };

  return addition;
}

```

do you really need the second function?

valerii-derkach (changes requested):

src/makeInfinityAdder.js
juliiaap: fixed

Virass (approved):

PR #725, opened 2022-08-13
Infinity adder created

stanislav-bzhezinskyi:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+  let sum = 0;
+
+  const getSum = (num) => {
+    if (num !== undefined) {
+      sum += num;
+
+      return getSum;
+    } else {
+      const result = sum;
+
```
valerii-derkach, line 18:
```suggestion
    } 
      const result = sum;

```

no need else statement after return

valerii-derkach (approved):

PR #753, opened 2022-08-22
adder working

Yevhen-Pavlovskyi:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
...
+      sum += x;
+
+      return adder;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    };
```
SanyaBratashchuk, line 23: You have return statement in if block so else is unnecessary

SanyaBratashchuk (approved):

PR #756, opened 2022-08-22
Infinity adder

Desertiks:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let summ = 0;
```
Reptiloid044, line 9: Better to rename  the variable correctly (smth. like "sum") It will be nicer to read

Reptiloid044 (changes requested):
Hi!
Please, rename the variable to make better code ;D

src/makeInfinityAdder.js
Desertiks: Ok, now its 'sum' 😄

Reptiloid044 (approved):
Thank you, now it looks nice!!!
Good job🚀

danylo-onishchuk (approved):

PR #758, opened 2022-08-22
Solution

plemiannik:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+
+  function adder() {
+    if (Object.values(arguments).length === 0) {
```
Alzay007, line 12: probably more comfortable without objectValues)))

Alzay007 (approved):
cool 🚀

src/makeInfinityAdder.js
plemiannik: the taste and color of all felt-tip pens are different

danylo-onishchuk (approved):

PR #766, opened 2022-08-23
solution

zinchenko-anastasiia:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
...
+
+      result = 0;
+
+      return temp;
+    }
+
+    return adder;
+  }
+
+  return adder;
```
VasylZinchenko, line 27: You can simplify your code a bit, deleting the second **if** and putting `return adder` into the first **if**

```suggestion
let result = 0;

  function adder(number) {
    if (typeof number !== 'undefined') {
      result += number;

      return adder;
    }

    const temp = result;

    result = 0;

    return temp;
  }

  return adder;
```

VasylZinchenko (approved):
Well done!👍

src/makeInfinityAdder.js
danylo-onishchuk: agree

danylo-onishchuk (changes requested):

danylo-onishchuk (approved):

PR #771, opened 2022-08-24
Solution of the task makeInfinityAdder

wfurs0w:


rboiar (approved):
Great solution 👍

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(number) {
```
AndriiRut, line 11: use arow function declaration

AndriiRut (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = number => {
```
AndriiRut, line 11:
```suggestion
  const adder = (number) => {
```

AndriiRut (approved):
🔥

PR #784, opened 2022-09-05
did the add function

HusakOleh:


wfurs0w (approved):
Good solution 👍🏻

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,29 @@
...
+      const res = adderInput.reduce((total, el) => {
+        return total + el;
+      }, 0);
+
+      adderInput = [0];
+
+      return res;
+    };
+
+    adderInput.push(n);
```
AndriiRut, line 22: if you'll just add the number to the adderInput you don't need to use reduce. reduce is overkill for this task)

AndriiRut (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,27 @@
...
+  };
+
+  return adder;
 }
 
 module.exports = makeInfinityAdder;
+
+// const adder = makeInfinityAdder();
+//
+// const actual = adder(1)(10)();
```
dellachaise, line 32: Delete comments.

dellachaise (approved):
Well done 👍

PR #789, opened 2022-09-14
Solution

MaxManis:


src/makeInfinityAdder.js
```diff
@@ -4,9 +4,36 @@
...
+// - Really wired way to calculate
+// |-- the sum of numbers, do you agree?))
 function makeInfinityAdder() {
-  // write code here
+  let total = 0;
+
+  function adder(x) {
+    total += x;
+
+    const infinityAdder = (y) => {
```
Denchepornyuk, line 15: you don't need infinityAdder function

move logic to adder function and remove infinityAdder
In adder you can check `x` type and return adder at the end of the function

Denchepornyuk (changes requested):

OleksiiBlezniuk (approved):

PR #794, opened 2022-09-17
Develop

grafit-off:
feat: makeInfinityAdder function was implemented

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let res = 0;
+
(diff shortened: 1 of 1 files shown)


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let res = 0;
```
TheVovchik, line 10: probably it's a short form of result word, don't looks like good naming

TheVovchik (approved):
names like res and sum don't sounds good
Great usage of recursion 👍

OleksiiBlezniuk (approved):
GJ!

PR #801, opened 2022-09-22
Hellish challenge

Vladosik1337:


jojnajbude (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      sumOfNUmbers += num;
+
+      return adder;
+    } else {
+      const finalValue = sumOfNUmbers;
+
+      sumOfNUmbers = 0;
+
+      return finalValue;
+    }
```
Anna-Tkachenko, line 22: else is redundant after return statement

Anna-Tkachenko (changes requested):

denys-cheporniuk (approved):

PR #805, opened 2022-09-22
Solution

Andrew-Skoryk:
infinityAdder was created

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  const adder = (value) => {
+    if (value === undefined) {
+      const temp = sum;
+
+      sum = 0;
+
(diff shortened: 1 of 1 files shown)


xiimiik (approved):
GJ 🔥

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  const adder = (value) => {
+    if (value === undefined) {
+      const temp = sum;
```
naziamoff, line 12: what do you mean by temp? Looks like you should fix this naming

naziamoff (approved):
Good job, but you should fix naming

src/makeInfinityAdder.js
Andrew-Skoryk: my idea was `tmp` = temporary

PR #807, opened 2022-09-22

xiimiik:


Andrew-Skoryk (approved):
GJ!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (num) => {
+    if (typeof num !== 'number') {
+      const finalSum = sum;
+
+      sum = 0;
```
Andr1yK, line 15: do u really need this?

Andr1yK (approved):

src/makeInfinityAdder.js
xiimiik: i think so...
When calling without parameters, it will return the result and clear the amount.

PR #814, opened 2022-09-23
function was implemented

ArturNagaycev:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,27 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (value) => {
+    if (value === undefined && sum === 0) {
+      return 0;
+    }
+
+    if (value === undefined && sum !== 0) {
```
glugovsky, line 16: You can combine these conditions. This way you will return `0` in case of sum being equal to `0`, and the `result` if the sum is not `0`
```suggestion
    if (value === undefined) {
```

ArturNagaycev: agree)) thanks! @glugovsky

glugovsky (approved):
👍

Anna-Tkachenko (approved):
🔥

PR #817, opened 2022-09-24
Solution

anastasiia-nahaiska:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  let sumIsShowed = 0;
```
myroslav-diiak, line 10: i think it can be boolean value instead of 1, 0

myroslav-diiak (approved):

src/makeInfinityAdder.js
anastasiia-nahaiska: Good point!))

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
...
+    if (sumIsShowed) {
+      sum = 0;
+      sumIsShowed = false;
+    }
+
+    if (arguments.length === 0) {
+      sumIsShowed = true;
+
+      return sum;
+    }
```
Denchepornyuk, line 22: don't use arguments use `number` indetead

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  let sumIsShowed = false;
```
Denchepornyuk, line 10: you don't need `sumIsShowed` variable, checking `number` argument in the function is enough

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #819, opened 2022-09-24
solution

artem-musii:


Dikey945:
Good job! 🔥

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  function adder(valueA) {
+    if (!arguments.length) {
+      return 0;
+    }
```
Denchepornyuk, line 12: don't use arguments, you have valueA and valueB, you them instead of arguments

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #822, opened 2022-09-25
solution

Naveed-Shoukat:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      const currentSum = sum;
+
+      sum = 0;
+
+      return currentSum;
+    }
+
+    sum += x;
+
+    return adder;
```
kamal-tufekcic, line 22: You can do inline return here

kamal-tufekcic (approved):

src/makeInfinityAdder.js
VinnieJ-2k20: Not sure how.

VinnieJ-2k20 (approved):
Good job 👍

src/makeInfinityAdder.js
VinnieJ-2k20: Oh, I see. Both options are fine.

PR #824, opened 2022-09-26
solution

vmamchur:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = number => {
+    if (number === undefined) {
+      const current = sum;
```
DenysKleimenov, line 13: I think a name like currentSum would be clearer

DenysKleimenov (approved):

src/makeInfinityAdder.js
Denchepornyuk: agree

Denchepornyuk (approved):

PR #830, opened 2022-09-28
-inifinite-sum-closures-

kolya-movchan:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
-}
+  let sum = 0;
+  let count = 0;
```
SanyaBratashchuk, line 10: Why do you need this count?

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
...
+      sum = 0;
+    }
+
+    if (number >= 0) {
+      sum += number;
+    } else {
+      count++;
+
+      return sum;
+    };
```
SanyaBratashchuk, line 23: Why do you need this check?

SanyaBratashchuk (changes requested):

src/makeInfinityAdder.js
kolya-movchan: It was created to check if there was an empty function call after an empty function call 
image

kolya-movchan: if there were previous calls and the function was return + the current call is not a number (in other words, empty call) then we will return 0 to reset

alexandra-protyanova (approved):

PR #838, opened 2022-10-01
solution

illia-kyselov:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+
+      sum = 0;
+
+      return resAdder;
+    } else {
+      sum += a;
+
+      return adder;
+    }
+  };
```
Virass, line 23: You don't need to use "else" operator here if you returned something on 17 lines.

Virass (changes requested):

AlisaVlasova (approved):

PR #843, opened 2022-10-03
solution

bradleymatjie:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let result = 0;
+
+  const add = function(number) {
+    if (arguments.length === 0) {
```
kamal-tufekcic, line 13: What if someone passes a string or `undefined`?

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let result = 0;
+
+  const add = function(number) {
```
kamal-tufekcic, line 12: Since it passed the tests I guess it's ok, but personally I'd either use an arrow function or function declaration

kamal-tufekcic (approved):

ihor-jpeg (approved):

PR #868, opened 2022-10-25
Solution

bokorzita:


src/makeInfinityAdder.js
```diff
@@ -5,8 +5,10 @@
  * @return {function}
  */
 
-function makeInfinityAdder() {
-  // write code here
+function makeInfinityAdder(a = 0) {
+  return (b) => {
+    return +b >= 0 ? makeInfinityAdder(a + b) : a;
```
jstmpelowycz, line 10: I believe you don't need type coercion for `b` 
```suggestion
    return b >= 0 
      ? makeInfinityAdder(a + b) 
      : a;
```
however, if you do need a variable to become `Number`, I suggest using `Number(...)` instead of `+...` – `+` operator is just too tiny to notice

jstmpelowycz (approved):

PR #869, opened 2022-10-25
hellish

metinbicaksiz:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,20 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  return function adder(value) {
+    if (value === undefined) {
```
jstmpelowycz, line 13:
```suggestion
    if (!value) {
```

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,20 @@
...
+      const newSum = sum;
+
+      sum = 0;
+
+      return newSum;
+    }
+    sum += value;
```
jstmpelowycz, line 20:
```suggestion
    }

    sum += value;
```

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,20 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  return function adder(value) {
+    if (value === undefined) {
+      const newSum = sum;
+
+      sum = 0;
```
jstmpelowycz, line 16: there is no need to do this step, I believe

jstmpelowycz (approved):

PR #877, opened 2022-10-31
solotion

MaxFedak:


Illia-Skladnik (approved):
I have nothing to add here.  Interesting solution!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,19 @@
...
 
 function makeInfinityAdder() {
-  // write code here
+  return function adder(num) {
+    if (typeof num === 'undefined') {
+      return 0;
+    }
+
+    return function(num2) {
+      if (num2 || num2 === 0) {
```
denys-danyliuk, line 15: is this check necessary?

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  return function adder(num) {
+    if (typeof num === 'undefined') {
+      return 0;
+    }
+
+    return function(num2) {
```
denys-danyliuk, line 14: try to figure it out without creating second function

denys-danyliuk (changes requested):

src/makeInfinityAdder.js
MaxFedak: I think so
we have 2 returns, and we need to check wen to return a func and when we return a value
if u mean zero check - without it solution crashes, cause 0 is falsy, but we need to keep going after it. So i made an exception for it

denys-danyliuk (approved):

PR #880, opened 2022-11-01
Made infinity adder

iryna-lelyk:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(number) {
```
masha-girya, line 11: wow, i see, that it could be wiht only one function

masha-girya (approved):
i have nothing to add, great job!

mykola-koval (approved):

PR #881, opened 2022-11-01
curry func MG

masha-girya:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,29 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  return (a) => {
+    let sum = a;
+
+    if (!a) {
+      return 0;
+    }
```
iryna-lelyk, line 14: You can remove this function and have only one function getSum inside makeInfinityAdder. Because in getSum  you return the function and your code will work fine without this one function.

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,29 @@
...
+
+    const getSum = (b) => {
+      if (b >= 0) {
```
iryna-lelyk, line 17: But what if we want to add negative numbers? Your function will just return total and reset sum to zero.

iryna-lelyk (approved):

src/makeInfinityAdder.js
masha-girya: yes, now I see, thanks!)

mykola-koval (approved):

PR #882, opened 2022-11-02
solution

taraselnik:


Anna-Shapovalova (approved):
Wow) It`s possible to use only one function inside) 🔥

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let total = 0;
+
+  return function adder(arg) {
+    const resultOnStop = total;
```
illiarerun, line 12: you only need it inside `if` block, so don't create redundant variables outside

illiarerun (approved):

PR #886, opened 2022-11-02
solution

Anna-Shapovalova:


Svitlana-Yudina (approved):
It's a very interesting solution! 🔥 I have another)) I could not understand how and where you accumulate the sum :eyes: There is no variable for this :upside_down_face:

PR #887, opened 2022-11-02
infinite sum

Svitlana-Yudina:


Anna-Shapovalova (approved):
Attagirl!) 👍

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  const adder = (num) => {
```
denys-danyliuk, line 10:
```suggestion
  let sum = 0;
  
  const adder = (num) => {
```
it's better to separate variables declaration and functions declaration just for better readability

denys-danyliuk (approved):

PR #888, opened 2022-11-03
complete

spiritkyiv:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let accamulate = 0;
```
Anna-Shapovalova, line 9: I think you`ve meant here accumulate (or accumulator?)

Anna-Shapovalova (approved):
Nice code) 👍

src/makeInfinityAdder.js
spiritkyiv: > I think you`ve meant here accumulate (or accumulator?)

Exactly)

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let accumulate = 0;
+
+  const f = (adder = null) => {
```
denys-danyliuk, line 11: Знімок екрана 2022-11-04 о 10 02 13

denys-danyliuk: also your func is taking number in parameters, you don't need to null it

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+    if (adder !== null) {
+      accumulate += adder;
+
+      return f;
+    } else {
```
denys-danyliuk, line 16: this else statement is redundant, code after return won't execute anyway

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      const answer = accumulate;
```
denys-danyliuk, line 17: find a better name

denys-danyliuk (changes requested):

illiarerun (approved):

PR #890, opened 2022-11-03
Task was Solved

Stepico:


andreyolin (approved):
good job!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(a) {
```
denys-danyliuk, line 11:
```suggestion
  return function adder(number) {
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+    const tmp = sum;
+
+    if (typeof a === 'undefined') {
```
denys-danyliuk, line 14:
```suggestion
    if (a === undefined) {
```
you can do this check even without typeof

denys-danyliuk (approved):

PR #893, opened 2022-11-03
solution

nazarko94:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result = 0;
+  const adder = (a = null) => {
```
Anna-Shapovalova, line 10: Are U sure U need to null the parameters?

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+    if (a === null) {
+      const add = result;
+
+      result = 0;
+
+      return add;
+    } else {
```
Anna-Shapovalova, line 17: I think U don`t need else here.

Anna-Shapovalova (approved):

Anna-Shapovalova (approved):
GJ) 👍  Left some comments)

nazarko94:
> GJ) 👍 Left some comments)

Thanks for review)

src/makeInfinityAdder.js
denys-danyliuk: there is no sense in doing that

src/makeInfinityAdder.js
denys-danyliuk: agree, it's redundant after return

src/makeInfinityAdder.js
denys-danyliuk, line 10:
```suggestion
  const adder = (number) => {
```

denys-danyliuk (changes requested):
Pay attention to the comments

PR #894, opened 2022-11-03
Solution

andrii-bassarab:


Largoxoxo (approved):
Nice work!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (num = null) => {
+    if (num === null) {
```
mykola-koval, line 12: you could've used undefined here, why you chose to use null?

mykola-koval (approved):

PR #896, opened 2022-11-03
Solution

andriy-shymkiv:


Anna-Shapovalova (approved):
Cool!) U use recursion) My first solution was the same) GJ) 👍

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  return function add(val) {
```
denys-danyliuk, line 9: if you return function right away you don't need to name it

denys-danyliuk (approved):

PR #900, opened 2022-11-04
Solution

referee1988:


Anna-Shapovalova (approved):
Good solution!) 👍

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(number) {
```
denys-danyliuk, line 11:
```suggestion
  return function (number) {
```
if you return function right away you can skip name

denys-danyliuk (approved):

PR #906, opened 2022-11-07
Solution

karyna-salnykova:


Yurii-Pozho:
I think your solution more better than my)   GJ!!!

illiarerun (approved):

PR #907, opened 2022-11-07
solution

Yurii-Pozho:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+  let sum = 0;
+
+  const adder = (num) => {
+    if (num === undefined) {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
```
Anna-Shapovalova, line 18: U don`t need else here)

Anna-Shapovalova (approved):
Good job!) Well done)

src/makeInfinityAdder.js
illiarerun: agree

illiarerun (changes requested):

illiarerun (approved):

PR #909, opened 2022-11-09
Infinity adder

maksym-kryvolap:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
...
+
+      return currentSum;
+    } else {
+      currentSum += x;
+
+      return sum;
+    }
+  }
+
+  return sum;
```
tetiana-ivanchenko, line 28:
```suggestion
  let currentSum = 0;

  function sum(x) {
    if (x === undefined) {
      const total = currentSum
      
      currentSum = 0;
      
      return total;
    } else {
      currentSum += x;

      return sum;
    }
  }

  return sum;
```

tetiana-ivanchenko (changes requested):

src/makeInfinityAdder.js
tetiana-ivanchenko: also work on naming - mb adder better name then sum for function?

tetiana-ivanchenko (approved):

PR #912, opened 2022-11-14
solution

euFarafonov:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function f(num) {
```
Bohdan-Maksymiuk, line 11:
```suggestion
  return function f(num) {
```
next time use better names then `f`

Bohdan-Maksymiuk (approved):

PR #913, opened 2022-11-15
infinityAdderDK

Kukharuk:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,28 @@
...
+
+      if (count === 1) {
+        return sum;
+      }
+
+      if (count > 1) {
+        sum = 0;
+
+        return sum;
+      }
```
loralevitska, line 28: DRY

```suggestion
      if (count > 1) {
        sum = 0;
      }
     
      return sum;
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,28 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let count = 0;
+  let sum = 0;
+
+  return function ad(x) {
+    if (x || x === 0) {
```
loralevitska, line 13: if you want to check number you can use  `typeof x === 'number'`

loralevitska (changes requested):
good! need some improvements

polosanya (approved):
Well done!

PR #918, opened 2022-11-16
solution

olenitut:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const add = (number) => {
+    if (number !== undefined) {
+      sum += number;
+
+      return add;
+    } else {
```
Viktor-Kost, line 16: You can avoid this additional `else` condition and write code below without else {} cause you return a value above

Viktor-Kost (changes requested):
All good, yet a little update waiting!)

Viktor-Kost (approved):
Fine!)

Virass (approved):

PR #919, opened 2022-11-17
Infinite sum solution

Sergey-Kovyazin:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function add(number) {
+    const bufer = sum;
+
+    if (arguments.length === 0) {
```
tetiana-ivanchenko, line 14:
```suggestion
    if (number === undefined) {
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function add(number) {
+    const bufer = sum;
```
tetiana-ivanchenko, line 12: you only need it inside if, so don't create redundant variables outside

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function add(number) {
```
tetiana-ivanchenko, line 11: maybe better to name fuction adder or smth else?

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #920, opened 2022-11-18
Implemented the solution for makeInfinityAdder function

zynkd:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+  return function adder(number) {
+    if (!arguments.length) {
+      const oldSum = sum;
+
+      sum = 0;
+
+      return oldSum;
+    }
+
+    sum += +number;
```
jstmpelowycz, line 20: I'd use `Number()` constructor, `+` is too tiny to notice
```suggestion
    sum += Number(number);
```

jstmpelowycz (approved):
LGTM

PR #926, opened 2022-11-20
solution

Roman-Malysh:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let res = 0;
+  let sum = 0;
+
+  const adder = (param) => {
+    if (param >= 0) {
```
tetiana-ivanchenko, line 14: actually, not a perfect solution, because if your parameter is -5, for example, your function will not work correctly. In your task, you do not have a condition that your parameters will always be greater than 0.

tetiana-ivanchenko (changes requested):

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,24 @@
 
 function makeInfinityAdder() {
   // write code here
+  let res = 0;
```
tetiana-ivanchenko, line 10: if you use your res only in the if block, better solution will be to declare it there

tetiana-ivanchenko, line 10:
```suggestion
```

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,24 @@
...
+  let sum = 0;
+
+  const adder = (param) => {
+    if (param === undefined) {
+      res = sum;
```
tetiana-ivanchenko, line 15:
```suggestion
      const res = sum;
```

tetiana-ivanchenko (changes requested):

src/makeInfinityAdder.js
Roman-Malysh: got it

nazarmatsevych (approved):

PR #928, opened 2022-11-21
Solution

Nazar-Brunarskyi:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(n) {
+    if (n === undefined) {
+      const sumToReturn = sum;
```
vadiimvooo, line 13: It would be better in this case if the variable were call "Number", not just n. But in general all looks well!

vadiimvooo (approved):

solaryasha (approved):
👍

PR #932, opened 2022-11-24
solution

Oleksandr-Filo:


shyptia (approved):
It looks goods. It is not very important, but it is said in the task `makeAdder function, which will return the adder function`

solaryasha (approved):
Good job!

PR #934, opened 2022-11-27
Solution

Vasyliy22:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
-}
+  let result = 0;
+
+  const adder = (a = null) => {
+    if (a === null) {
```
Bohdan-Maksymiuk, line 12:
```suggestion
  const adder = (a) => {
    if (a === undefined) {
```

Bohdan-Maksymiuk: do you really need `null` as a default value?

Bohdan-Maksymiuk (approved):

PR #937, opened 2022-11-28
Solution

webdevw:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+  let result;
+
+  function adder(number) {
+    if (number === undefined) {
+      result = amount;
```
tetiana-ivanchenko, line 14:
```suggestion
      const result = amount;
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+  let result;
```
tetiana-ivanchenko, line 10:
```suggestion
```

src/makeInfinityAdder.js
tetiana-ivanchenko: if you use your variable only in this block, create it here

tetiana-ivanchenko (changes requested):

polosanya (approved):
Well done!

PR #949, opened 2022-12-02
solution

nikachu404:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (number === undefined) {
+      const total = sum;
+
+      sum = 0;
+
+      return total;
```
anastasiia-bilska, line 17: Good but maybe you should name whis variables more understandable

anastasiia-bilska (changes requested):
👍

solaryasha (approved):
👍

PR #951, opened 2022-12-02
solution

lefteris-L:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let tempSum = 0;
+  let sum;
+
+  return function add(n) {
+    if (isNaN(n)) {
```
ysrckr, line 13: I liked this approach. You can add `&& n === undefined` so `adder(2)('str')` doesn't return a value.

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let tempSum = 0;
+  let sum;
```
ysrckr, line 10: you can set `let sum = 0` and inside the function you can set `
let tempSum  =  sum
sum = 0
return tempSum
`
with this you can have only one global variable.

ysrckr (approved):
Just some suggestion but everything looks ok.

meforoma (approved):
working solution 👍

PR #962, opened 2022-12-07
done

IoghiStudio:


CharAnagn (approved):
LGTM! 👍

src/makeInfinityAdder.js
```diff
@@ -1,12 +1,23 @@
...
+  let result = 0;
+
+  const add = a => {
+    if (a === undefined) {
+      const output = result;
+
+      result = 0;
+
+      return output;
+    }
```
meforoma, line 13: just provide a default value for a - instead of checking for undefined
```suggestion
  const add = (a = 0) => {
    ...
```

meforoma (changes requested):
unnecessary check for `undefined`

src/makeInfinityAdder.js
meforoma: @IoghiStudio to address your concern - if your code passes tests, you can be approved.
If you don't agree with review comments, you can tell us why.
If you want to get approved - as said earlier. You can be.

meforoma (approved):

PR #971, opened 2022-12-08
solution

Lun4trick:


IoghiStudio (approved):
Look good!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+
+      result = 0;
+
+      return saveValue;
+    }
+
+    result += num;
+
+    return add;
+  };
```
meforoma, line 23: use anonymous / arrow function
don't check for undefined - but provide an initial value instead
```suggestion
  return add = (num = 0) => {
    ...
  };
```

meforoma (changes requested):
please see the comment

naziamoff (approved):

PR #974, opened 2022-12-08
Solution

CharAnagn:


ysrckr (approved):
Looks good

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,21 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  return function adder(num) {
+    if (num === undefined) {
+      const innerSum = sum;
```
meforoma, line 14:
```suggestion
  return function adder(num = 0) {
    if (num === undefined) {//would be easier to provide a default value
      const innerSum = sum;
```

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,21 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  return function adder(num) {
```
meforoma, line 12: it's a common practice to return an anonymous / array function
since declaring with `function` special word hoists the function - you don't need it
```suggestion
  return (num = 0) => {
```

meforoma (changes requested):
please see few comments

CharAnagn:
@meforoma Hey Roman, I understand your comments, and I tried a few solutions but it's fails at least one test everytime 😕

naziamoff (approved):

PR #975, opened 2022-12-08
Infinite sum closures

OlexandraKorol:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (value) => {
```
VladimirEdelmann, line 11: Var name `number` would be clearer to me

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+    const setValue = sum;
+
+    if (!value && value !== 0) {
```
VladimirEdelmann, line 14: `if (number === undefined) {`

VladimirEdelmann (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (value) => {
+    const setValue = sum;
```
VladimirEdelmann, line 12: lastValue is clearer

VladimirEdelmann (changes requested):

src/makeInfinityAdder.js
OlexandraKorol: great point!

valerii-derkach (approved):

PR #978, opened 2022-12-08
Solution

4ebupel:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+
+  const add = (num) => {
```
oleh-yablunovskyi, line 11:
```suggestion
  const adder = (num) => {
```
If your function intended to make adder, maybe it's better to return exactly 'adder' from it.

oleh-yablunovskyi (approved):
Good job!

PR #979, opened 2022-12-08
solution

mmahfuzi:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,21 @@
...
+
+      sum = 0;
+
+      return result;
+    }
+
+    sum += x;
+
+    return adder;
+  };
```
meforoma, line 24: better use anonymous function
and no need to check for `undefined` - just provide an initial value
```suggestion
  return adder = (x = 0) {
    ...
  };
```

meforoma (changes requested):
please see the comment

mmahfuzi:
> please see the comment

Hello @meforoma thanks for the comments. I tried to follow your suggestion but I got many errors! My code is not acceptable as it is?

src/makeInfinityAdder.js
meforoma: i forgot an actual arrow for an arrow function
did you use it?
`return adder = (x = 0) => {`
weird that you've got lots or errors.

Yet we could keep your code as is.

meforoma (approved):

src/makeInfinityAdder.js
mmahfuzi: Yes I added the => but didn't work! 

Thank you :)

PR #983, opened 2022-12-08
solution

Sveta-Kryukova:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(x) {
```
Oleksandr-Filo, line 11: you should use more understandable names for variables (not 'x')

Oleksandr-Filo (approved):
looks good, just one thing

valerii-derkach (approved):

PR #985, opened 2022-12-08
solution

lukacsja:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  let result;
+
+  const adder = (n) => {
+    if (n === undefined) {
```
naziamoff, line 13: do you really need to check for undefined, or you're interested in whether the value is falsy?

naziamoff: if it is falsy, you can just provide default value there

naziamoff (changes requested):

src/makeInfinityAdder.js
lukacsja: I spent a lot of time with it and could not manage to pass tests without checking for undefined or argument.length. Maybe I misunderstood idk. May I have your suggestion please?

naziamoff: I meant checking like `!n`
But probably I'm not aware of all the test of this task

naziamoff (approved):

src/makeInfinityAdder.js
lukacsja: oh yeah thats doable, I overthought a bit. thank you!

PR #990, opened 2022-12-09
Solution

viktoriia-kravchenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  return function adder(a) {
+    if (a >= 0) {
```
olena-yanovska, line 10: I'm not sure it's a good idea to compare a and b with 0. This function will not work with negative numbers. You can just check if they are not undefined

olena-yanovska (changes requested):

src/makeInfinityAdder.js
viktoriia-kravchenko: Good point. Thanks!

olena-yanovska (approved):

valerii-derkach (approved):

PR #1002, opened 2022-12-10
solution

olena-yanovska:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(value) {
+    const sumResult = sum;
```
VladimirEdelmann, line 12: lastValue or lastResult or lastSum would be clearer to me

Ihor-Karpyn: Agree

Ihor-Karpyn (approved):
GJ, pay more attention to variable naming

src/makeInfinityAdder.js
olena-yanovska: fixed

PR #1018, opened 2022-12-15
Solution for make infinity adder

katerinalex:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,44 @@
...
+
+      if (num === undefined && count === 1) {
+        currentSum = 0;
+
+        return currentSum;
+      }
+      currentSum += num;
+
+      return result;
+    };
```
loralevitska, line 41: avoid code duplicate. Think about optimize. If you need you can use helper functions or common constants

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,38 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let currentSum = 0;
+  let count = 0;
```
Anastasiia-Svintsova, line 10: You don't need two variables here. 
```suggestion
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,38 @@
...
+  const adder = (number) => {
+    if (!number) {
+      currentSum = 0;
+
+      return currentSum;
+    }
```
Anastasiia-Svintsova, line 16:
```suggestion
    if (!number) {
      const result = currentSum;
      currentSum = 0;

      return result;
    }
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,38 @@
...
+        return currentSum;
+      }
+      currentSum += num;
+
+      return result;
+    };
+
+    result(number);
+
+    return result;
```
Anastasiia-Svintsova, line 37: You don't need this large construction
```suggestion
    currentSum += number;
    
    return adder
```

Anastasiia-Svintsova (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let currentSum = 0;
+  const adder = (number) => {
```
polosanya, line 10:
```suggestion
  let currentSum = 0;
  
  const adder = (number) => {
```

polosanya (approved):
Good job!

PR #1020, opened 2022-12-20
Solution

IrynaK-a:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,24 @@
...
+
+      return iterate;
+    } else {
+      const allNumbers = [...numbers];
+      const endResult = allNumbers.reduce((sum, number) => sum + number, 0);
+
+      numbers = [];
+
+      return endResult;
+    }
```
loralevitska, line 23: try to simplify the solution. `reduce ` means each new loop. You know that the parameters will be numbers, so do it with the help of numbers. `const count = 0` ... `count +=  ...`

loralevitska (changes requested):
almost perfect

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let count = 0;
+
+  const iterate = (figure) => {
+    if (figure !== undefined) {
+      count += figure;
+
+      return iterate;
+    } else {
```
Anastasiia-Svintsova, line 16: Redundant else statement here. You have a return above 
```suggestion
    }
```

Anastasiia-Svintsova (approved):

nazarmatsevych (approved):

PR #1027, opened 2022-12-29
solution

Malva37:


OlhaYevstifieieva (approved):
Looks great! 👌🏻

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+  let sum = 0;
+
+  const adder = (a) => {
+    if (a === undefined) {
+      const tmp = sum;
+
+      sum = 0;
+
+      return tmp;
+    } else {
```
alexandra-protyanova, line 18: you can remove else here

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (a) => {
+    if (a === undefined) {
```
```suggestion
    if (!a) {
```

alexandra-protyanova (approved):

src/makeInfinityAdder.js
Malva37: tried to implement your advice but met with next (arg 0 converted to false)  https://prnt.sc/m6zj1hB4LKJa

PR #1029, opened 2023-01-01
Solution

NVyrsta:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,20 @@
...
+    if (num === undefined) {
+      const total = sum;
+      sum = 0;
+
+      return total;
+    }
+
+    sum += num;
+
+    return add;
```
OlhaYevstifieieva, line 21: Do you need return Add function twice inside this function?

OlhaYevstifieieva (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function add(num) {
+    if (num === undefined) {
```
```suggestion
    if (!num) {
```

alexandra-protyanova (approved):

PR #1031, opened 2023-01-02
Solution

dissolveyo:


DarkMistyRoom (approved):
It works, but no need to store an array of all numbers. You improve performance by storing only the sum

PR #1032, opened 2023-01-02
infiniteAdder task completed

MykolaKaradzha:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let counter = 0;
+  let finalSum;
+
+  const infiniteAdder = (...args) => {
```
Viktor-Kost, line 12: Is it gonna be more than one argument? 
So if not - let function takes only one argument

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+    if (args.length) {
+      counter += args[0];
```
Viktor-Kost, line 14:
```suggestion
    if ({{your_argument}} !== undefined) {
      counter += {{your_argument}} ;
```

Viktor-Kost (approved):
Good job!
Make some improvements for yourself

src/makeInfinityAdder.js
MykolaKaradzha: Got it, thanks)

PR #1041, opened 2023-01-05
add solution

Anastasiia-Khudych:


OlhaYevstifieieva (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function add(x) {
+    if (x === undefined) {
```
```suggestion
    if (!x) {
```

alexandra-protyanova (approved):

src/makeInfinityAdder.js
Anastasiia-Khudych: It will not work with zero as an argument.

PR #1045, opened 2023-01-06
solution

igor-kashchenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(num) {
+    if (num !== 0 && !num) {
```
vanst1, line 12: you can use `num === undefined`

stadnytskyi-ivan (approved):
Well done;)

src/makeInfinityAdder.js
alexandra-protyanova: agree

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      const sums = sum;
```
alexandra-protyanova, line 13: bad name for variable

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1052, opened 2023-01-08
Solution

sv-dubov:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,24 @@
...
 function makeInfinityAdder() {
-  // write code here
+  return (firstNumber) => {
+    let sum = firstNumber;
+
+    if (firstNumber === undefined) {
+      sum = 0;
+
+      return sum;
+    }
```
GarryRocksman, line 16: So here you check if the number is undefined, and you do the same check below in add function. I think it's better to use one function to return and check input values once

GarryRocksman (approved):
GJ,🔥 but I think needs some modification

src/makeInfinityAdder.js
sv-dubov: Thanks, remaked

Denchepornyuk (approved):

PR #1056, opened 2023-01-09
Task was solved

ihor-soloviov:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
-  // write code here
+  let storage = [];
+
+  const addition = (value) => {
+    if (value >= 0) {
+      storage.push(value);
+
+      return addition;
+    } else {
+      const res = storage.reduce((acc, item) => acc + item, 0);
```
Avichez, line 17: Try to give full names for variables.

Avichez (approved):
Great solution 🔥

src/makeInfinityAdder.js
alexandra-protyanova: agree

alexandra-protyanova (approved):

PR #1059, opened 2023-01-10
task done

Bogdan-Kutasevich:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  const adder = (...number) => {
+    if (number.length) {
+      sum += number[0];
```
vanst1, line 12:
```suggestion
  const adder = (number) => {
    if (number !== undefined) {
      sum += number;
```

vanst1 (changes requested):
Good job:)
some fix needed)

src/makeInfinityAdder.js
Bogdan-Kutasevich: thanks for advice)

alexandra-protyanova (approved):

PR #1065, opened 2023-01-12
v-01

AndreyBerezyanskiy:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+  if (x !== undefined) {
+    sum += x;
+  } else {
+    sum = 0;
+  };
+
+  return x === undefined
+    ? tempSum
+    : adder;
+}
```
oleh-holovnykh, line 23: two similar checks are redundant, i think
its better to `return tempSum;` in else statement and delete ternary

oleh-holovnykh (changes requested):
Good job! just one suggestion

src/makeInfinityAdder.js
oleh-holovnykh: if x !== undefined, we increase sum and return function. Else reset sum, and return tempSum

AndreyBerezyanskiy: I'm a little bit confused) I didn't see this nice solution. Thanks!)

Denchepornyuk (approved):

PR #1068, opened 2023-01-13
Solution

Coyoote:


Avichez (approved):
Great and clear solution 🔥

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let numbers = 0;
+
+  return function adder(num) {
+    if (typeof num === 'number') {
+      numbers += num;
+
+      return adder;
+    } else {
```
Anna-Tkachenko, line 16: else is redundant here
if code inside `if` statement will be executed, your function will be ended with return statement

Anna-Tkachenko (approved):

src/makeInfinityAdder.js
Coyoote: Thanks for the comment, removing it right away!

PR #1074, opened 2023-01-16
Solution

natalia-maryshko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (a) => {
+    if (a >= 0) {
```
```suggestion
    if (a !== undefined) {
```
I think it's more correct to check if it's not undefined because closure should stop if there are no parameters (a === undefined)

polosanya (approved):

PR #1077, opened 2023-01-22
Solution

ruslan-yarosh:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,17 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+
+  const sum = (num) => {
+    const result = amount;
+
+    amount = num === undefined ? 0 : amount + num;
+
+    return num === undefined ? result : sum;
```
DarkMistyRoom, line 16: you can use if to avoid repeating

DarkMistyRoom (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
 
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+
+  const sum = (num) => {
+    const result = amount;
+
+    if (num === undefined) {
+      amount = 0;
```
tetiana-ivanchenko, line 15: move the creation of the result variable to the "if" condition

tetiana-ivanchenko (changes requested):

SanyaBratashchuk (approved):

PR #1079, opened 2023-01-23
Infinity Adder

Rio1391:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(number) {
+    if (number === parseInt(number)) {
```
DarkMistyRoom, line 12: It will sum all the numbers passed in the parameters until it encounters a call without parameters. So would be better to check if `number === undefined` because now it will work the same with any other type except number, for example, some string, but this is not what is required in the condition

DarkMistyRoom (changes requested):

loralevitska (approved):

PR #1080, opened 2023-01-23
infinity adder was made

StetsenkoViktoriia:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (args) => {
```
Anastasiia-Svintsova, line 11: You don't have a lot of arguments here, just one. So better to use another name, for example
```suggestion
  const adder = (value) => {
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
+      sum += args;
+
+      return adder;
+    }
```
Anastasiia-Svintsova, line 22: Redundant else statement here. You have a return before
```suggestion
      return result;
    } 
    
    sum += args;

    return adder;
```

Anastasiia-Svintsova (changes requested):

polosanya (approved):
Good job!

PR #1089, opened 2023-02-01
sollution

serbinm:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      const answer = sum;
+
+      sum = 0;
+
+      return answer;
+    } else {
+      sum += a;
+
+      return add;
+    }
```
polosanya, line 22:
```suggestion
      return answer;
    }
    
    sum += a;

    return add;
```
No need for else. You code will not go further after return anyway

polosanya (approved):
Good job!

PR #1090, opened 2023-02-01
Chupiak hw

Chupiak:


src/makeInfinityAdder.js
```diff
@@ -4,9 +4,25 @@
  *
  * @return {function}
  */
-
 function makeInfinityAdder() {
-  // write code here
+  let total = 0;
+
+  // eslint-disable-next-line no-shadow
```
DarkMistyRoom, line 10: why do you need it? make it work without

src/makeInfinityAdder.js
```diff
@@ -4,9 +4,25 @@
...
+  function adder(item) {
+    if (item !== undefined) {
+      total = total + item;
+    } else {
+      const result = total;
+
+      total = 0;
+
+      return result;
+    }
```
DarkMistyRoom, line 20: refactor to avoid `else`. you can do the same with only one plane `if`

src/makeInfinityAdder.js
```diff
@@ -4,9 +4,25 @@
...
  */
-
 function makeInfinityAdder() {
-  // write code here
+  let total = 0;
+
+  // eslint-disable-next-line no-shadow
+  function adder(item) {
+    if (item !== undefined) {
+      total = total + item;
```
DarkMistyRoom, line 13:
```suggestion
      total += item;
```

DarkMistyRoom (changes requested):

SanyaBratashchuk:
Why did you close this PR?

PR #1095, opened 2023-02-03
Adder was created

Vasyl-Pavlenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result = 0;
+  let sum = 0;
```
witflash, line 10: You can remove this const here...

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+  const adder = (a = null) => {
+    if (a === null) {
+      sum = result;
```
witflash, line 13: ... and use just here: `const sum = result;`

So you need this `sum` only inside this block.

witflash (approved):
Good job. I've left some comments for your future work

nazarmatsevych (approved):

PR #1110, opened 2023-02-11
solution

xnquaint:


VitaliyBondarenko1982 (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const add = (x) => {
```
Volodymyr-Melnyk-Fron10der, line 11:
```suggestion
  const add = (number) => {
```

Volodymyr-Melnyk-Fron10der (approved):
Great, but try to make more meaningful names on for next time

PR #1113, opened 2023-02-13
Solution Infinite sum closures

Sasha-Green:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+
+  const adder = arg => {
+    if (arg === undefined) {
```
OleziO, line 12: Use `!arg`, its the same as `arg === undefined`

OleziO (approved):

src/makeInfinityAdder.js
Sasha-Green: I tried but that option didn't work when the argumen is zero

PR #1115, opened 2023-02-14
Solution

vlad-ptashnichenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      const count = amount;
+
+      amount = 0;
+
+      return count;
+    } else {
+      amount += x;
+
+      return adder;
+    }
```
Viktor-Kost, line 22: you have `return` for your if condition, so in this case you don't need to use `else`
```suggestion
    } 
    
      amount += x;

      return adder;
   
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+
+  return function adder(x) {
+    if (!x && x !== 0) {
```
Viktor-Kost, line 12:
```suggestion
    if (x === undefined) {
```

Viktor-Kost (changes requested):
Nice!
Just a few moments

VitaliyBondarenko1982 (approved):

PR #1120, opened 2023-02-16
solution

AntonKabanets:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (number === undefined) {
```
OleziO, line 12: Replace with `!number`, it's the same as `number === undefined`

OleziO (approved):

OleziO (changes requested):
Fix test error.

PR #1125, opened 2023-02-20
Solution

nastychenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (...args) => {
+    if (args.length < 1) {
```
sTorba24, line 12:
```suggestion
    if (num === undeifined) {
```

sTorba24 (changes requested):
Well done! Little comments to fix

qoa11a (approved):

PR #1132, opened 2023-02-22
commit message

Peti51:


src/makeInfinityAdder.js
```diff
@@ -5,8 +5,23 @@
...
  */
 
-function makeInfinityAdder() {
-  // write code here
+function makeInfinityAdder(a = 0) {
+  let sum = 0;
+  let secondSum;
+  const makeAdder = (b) => {
+    if (b === undefined) {
+      secondSum = sum;
```
polosanya, line 13:
```suggestion
  let sum = 0;

  const makeAdder = (b) => {
    if (b === undefined) {
      const finalSum = sum;
```
You just create it only inside second function.

polosanya (approved):

PR #1133, opened 2023-02-22
js_sum-with-closure

JulyaPetrovskaya:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  function adder(n) {
```
Viktor-Kost, line 10: It's a good practice to apply more clear names
```suggestion
  function adder(number) {
```

Viktor-Kost (approved):
Good job!

PR #1139, opened 2023-02-24
infinite adder was created

rashyd-hasratov:


NikitaStambul (approved):
great job

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const add = (a) => {
```
sergii-nosachenko, line 11: What is `a`? Maybe: `value`, `operand`...

sergii-nosachenko (changes requested):
Quick fix

gretoq (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+
+      return final;
+    }
+
+    sum += operand;
+
+    return add;
+  };
+
+  return add;
```
illiarerun, line 25: you can return function right away
```suggestion
  return (operand) => {
    if (operand === undefined) {
      const final = sum;

      sum = 0;

      return final;
    }

    sum += operand;

    return add;
  };;
```

illiarerun (approved):

PR #1143, opened 2023-02-24
makeInfinityAdder was implement

KuharchukDmytro:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(input) {
+    if (!input && input !== 0) {
```
MaksymMaliuk, line 12: I think it's better to do it this way
```suggestion
    if (input === undefined) {
```

sergii-nosachenko (approved):
Good work 👍

PR #1144, opened 2023-02-25
Solution for Make Infinity Adder

ctk12:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let total = 0;
+
+  const add = (value) => {
```
qoa11a, line 11:
```suggestion
  const adder = (value) => {
```

qoa11a (approved):
Good job!

PR #1145, opened 2023-02-26
Infinite Sum Closures - Solution

Yevhenii0536:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
...
+      sum += number;
+
+      return adder;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
KuharchukDmytro, line 23: you dont need else here, if first if statement isnt true, next block will run anyway

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
...
+  }
+
+  return adder;
```
KuharchukDmytro, line 26:
```suggestion
  return (number) => {
    if (number !== undefined) {
      sum += number;

      return adder;
    } else {
      const result = sum;

      sum = 0;

      return result;
    }
  }
```

KuharchukDmytro: if you return some function, you can do it this way and simplify your code

KuharchukDmytro (changes requested):
good job, but there is some little mistakes, fix them please:3

KuharchukDmytro (approved):
gj!

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
```
sergii-nosachenko, line 9: Cleanup code

sergii-nosachenko (approved):
Good work 👍

PR #1148, opened 2023-02-27
Created chained function for sum counting

gretoq:


Yekatery (approved):
Well done!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (...args) => {
```
illiarerun, line 11: it's not in the requirements, but we don't really need to accept more than one argument, next time do what is required, sometimes doing more !== better

illiarerun (approved):

PR #1153, opened 2023-02-28
Solution

BozhenaBodnar:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(num) {
```
illiarerun, line 11: okay, it's needed here to return function inside itself but you really don't want to have a function name when you're returning it

illiarerun (approved):

PR #1160, opened 2023-03-02
Solution

DaveBeetle:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+  let numbers = [];
+
+  const adder = number => {
+    if (number !== undefined) {
+      numbers.push(number);
+
+      return adder;
+    }
+
+    const result = numbers.reduce((a, b) => (a + b), 0);
```
SerhiiKirik, line 18: will be better for the algorithm, if you just create counter & just sum these numbers

SerhiiKirik (approved):
GJ 🔥

anstsot (changes requested):
I found one issue. Let's fix your code
I'm trying to call `adder(5)(5)(5)` and it has to return `15` but it's not working correctly. could you please check this case?

anstsot (approved):

PR #1161, opened 2023-03-02
'Solution'

nazarkharkevych:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+  let sum = 0;
+
+  const adder = (...args) => {
+    const returnValue = sum;
+
+    if (!args.length) {
+      sum = 0;
+
+      return returnValue;
+    }
```
Volodymyr-Melnyk-Fron10der, line 18: I would move the variable  under if block, 'cause it's being used only there
```suggestion
  const adder = (...args) => {
    if (!args.length) {
      const returnValue = sum;
      sum = 0;

      return returnValue;
    }
```

Volodymyr-Melnyk-Fron10der (approved):
Good job✅ check only one not critical recommendation

PR #1175, opened 2023-03-07
solution

serhiilysenko:


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.9.7",
+    "@mate-academy/scripts": "^1.2.8",
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      res = 0;
+
+      return totalRes;
+    }
+
+    res += val;
+
+    return add;
+  };
 }
```
SerhiiKirik, line 24: Alternative solution
```suggestion
  const add = (value) => {
    if (value === undefined) {
      const totalRes = res;

      res = 0;

      return totalRes;
    }

    res += value;

    return add;
  }
  
  return add;
}
```

SerhiiKirik (approved):
Great 🔥

PR #1176, opened 2023-03-08
'done'

Hanna-Balabukha:


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.9.7",
+    "@mate-academy/scripts": "^1.2.8",
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,27 @@
...
+
+        result = 0;
+
+        return log;
+      } else {
+        result += number;
+
+        return adder;
+      }
+    };
```
SerhiiKirik, line 24: It's extra, bc you will have false in 13 lines - the code will continue
```suggestion
      }
      
      result += number;

      return adder;
    };
```

SerhiiKirik (approved):
Well done 🚀

PR #1181, opened 2023-03-09
complete task

VladGraund:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let emptyAgain = 0;
+  let sum = 0;
+  const addAllCurry = (a) => {
+    if (a !== undefined) {
```
OleziO, line 12: You can remove this condition, just add the default value (like zero) to `a` param.

OleziO: Rename `a` to more meaningful name.

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let emptyAgain = 0;
```
OleziO, line 9: You don't need this variable, try to refactor without it. If you have any questions - feel free to ask them in `fe_chat`.

OleziO (changes requested):

PR #1182, opened 2023-03-10
js_task-inifinite-sum-closures task done

EhorSaprykin:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,19 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
-}
+  return function add(x) {
+    if (x || x === 0) {
```
polosanya, line 10:
```suggestion
    if (x !== undefined) {
```
Won't it work the same?

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,19 @@
...
+      return function(y) {
+        if (y || y === 0) {
+          return add(x + y);
+        } else {
+          return x;
```
polosanya, line 15: You can use ternary operator here to make your code a bit cleaner

polosanya (changes requested):
Let's improve your code a bit)

nazarmatsevych (approved):
👍

PR #1186, opened 2023-03-13
Solution

oksanatytanych:


src/makeInfinityAdder.js
```diff
@@ -5,8 +5,24 @@
  * @return {function}
  */
 
-function makeInfinityAdder() {
-  // write code here
+function makeInfinityAdder(n) {
+  let result = 0;
+  let zm = 0;
```
IvanFesenko, line 10: Not clear what does zm mean?

src/makeInfinityAdder.js
```diff
@@ -5,8 +5,24 @@
...
+    if (value !== undefined) {
+      zm += value;
+
+      return add;
+    } else {
+      result = zm;
+      zm = 0;
+
+      return result;
+    }
```
IvanFesenko, line 22: Else statement is useless here
```suggestion
    } 
    
    result = zm;
    zm = 0;

    return result;
```

IvanFesenko (changes requested):
good job. Fix naming and remove else statement

src/makeInfinityAdder.js
```diff
@@ -5,8 +5,24 @@
  * @return {function}
  */
 
-function makeInfinityAdder() {
-  // write code here
+function makeInfinityAdder(n) {
+  let result = 0;
```
etojeDenys, line 9: you can move this line into `add` function

etojeDenys (approved):
good job

PR #1187, opened 2023-03-13
Solved

silcki:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,26 @@
 
 function makeInfinityAdder() {
   // write code here
+  const arr = [];
+
+  const f = (x) => {
```
Viktor-Kost, line 12: it's better to give meaningful names for your variables

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,26 @@
 
 function makeInfinityAdder() {
   // write code here
+  const arr = [];
```
Viktor-Kost, line 10: You have a nice solution, yet you can simplify your code and avoid array with reduce methods
```suggestion
let result = 0;
```

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,26 @@
...
+        (accumulator, currentValue) => accumulator + currentValue,
+        0
+      );
+
+      arr.length = 0;
+
+      return summ;
+    }
+
+    arr.push(x);
```
Viktor-Kost, line 24:
```suggestion
result += value;
```

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,26 @@
...
+  const f = (x) => {
+    if (x === undefined) {
+      const summ = arr.reduce(
+        (accumulator, currentValue) => accumulator + currentValue,
+        0
+      );
+
+      arr.length = 0;
+
+      return summ;
```
Viktor-Kost, line 21:
```suggestion
      const sum = result;
      result = 0;
      return sum;
```

Viktor-Kost (changes requested):

PR #1190, opened 2023-03-14
Solution

dudar-y:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
...
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function(x) {
+    if (x === undefined) {
+      return 0;
+    }
+    sum += x;
```
DarkMistyRoom, line 15: this function should return current sum(not 0) when x is not passed and set 0 to sum

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
...
+
+    const adder = function(y) {
+      if (y !== undefined) {
+        sum += y;
+
+        return adder;
+      }
+
+      return sum;
+    };
```
DarkMistyRoom, line 25: you don't need this function. it makes logic too complicated. your first function(function(x)) can be the adder and return itself

DarkMistyRoom (changes requested):

Volodymyr-Melnyk-Fron10der (approved):
Perfect now!

PR #1194, opened 2023-03-17
Completed task

metja:
I add in IF operator OR because 0 its false

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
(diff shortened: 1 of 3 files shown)


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(number) {
+    if (number || number === 0) {
```
SanyaBratashchuk, line 12:
```suggestion
    if (number !== undefined) {
```

SanyaBratashchuk (approved):

PR #1196, opened 2023-03-20
Solution

LBondarchuk:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,34 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
-}
+  let sum = 0;
+  const functionA = numberA => {
+    if (numberA === undefined) {
```
OleziO, line 11: Replace with `!numberA`, it will be the same as` numberA === undefined`

Rename `numberA` to `a`

OleziO: Same situation with `numberB`

OleziO (changes requested):

polosanya (approved):
I'm approving it, but please pay attention to the comment)

PR #1198, opened 2023-03-21
Probably not Solution

DanielSorokowski:
I know the code is hell but I have no idea how to loop it.

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -9,7 +9,7 @@ function makeInfinityAdder() {
   let sum = 0;
 
   const adder = (a) => {
-    if (!a && a !== 0) {
+    if (typeof a === 'undefined') {
       const result = sum;
 
       sum = 0;


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,69 @@
...
 
 function makeInfinityAdder() {
-  // write code here
+  const adder = (a) => {
+    if (!a && a !== 0) {
+      return 0;
+    }
+
+    return (b) => {
+      if (!b && b !== 0) {
```
solaryasha, line 15: You should use `closure` and store result of the sum;

Function `adder` should return itself if we pass the param. If not - we should return the counter

solaryasha (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (a) => {
+    if (!a && a !== 0) {
```
solaryasha, line 12: Let's use `typeof` operator and check if the a is `undefined`

solaryasha (changes requested):

solaryasha (approved):

PR #1201, opened 2023-03-22
Solution

igorkruz:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+
+  const adder = (value) => {
+    if (value === undefined) {
+      const count = amount;
+
+      amount = 0;
+
+      return count;
+    } else {
+      amount += value;
```
PavloReutskiy, line 19: I hear the mentor's voice here: 'value is a bad name, son'... 'may the force be with you' 😎

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+
+  const adder = (value) => {
+    if (value === undefined) {
+      const count = amount;
```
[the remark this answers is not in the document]
igorkruz: I think it unnecessary to create a variable outside the if because I use this variable only in this if statement.

src/makeInfinityAdder.js
igorkruz: fixed )

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.9.7",
+    "@mate-academy/scripts": "^1.2.8",
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+
+      amount = 0;
+
+      return count;
+    } else {
+      amount += number;
+
+      return adder;
+    }
```
SerhiiKirik, line 22: You don't need else, bc, if number === undefined - you make return from your function
```suggestion
    }
    
    amount += number;

    return adder;
    }
```

SerhiiKirik (approved):
Great 🚀

PR #1207, opened 2023-03-25
solution

glots-a:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function cyclingFunc(param) {
```
BondarenkoDmitriy, line 11: you can use '...args' instead of 'param' and and check if parameters are passed using if (args.length > 0).
I think this is not a mistake, just a more general way

BondarenkoDmitriy (approved):
GJ

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+    if (param !== undefined) {
+      sum += param;
+
+      return cyclingFunc;
+    } else {
```
SerhiiKirik, line 16: you don't need else here
```suggestion
    }
```

SerhiiKirik (approved):
Well done 🚀

PR #1209, opened 2023-03-26
function works

BondarenkoDmitriy:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(...args) {
+    if (args.length > 0) {
+      sum += args.reduce((acc, val) => acc + val, 0);
+
+      return adder;
+    } else {
```
Esteban4769, line 16: Dont really need else here

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(...args) {
+    if (args.length > 0) {
+      sum += args.reduce((acc, val) => acc + val, 0);
```
Esteban4769, line 13: Not sure you need to handle the case with multiple args, but its okay.

Esteban4769 (approved):
Good job

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.9.7",
+    "@mate-academy/scripts": "^1.2.8",
```

src/makeInfinityAdder.js
SerhiiKirik: agree

SerhiiKirik (changes requested):
Well done 🚀 
Something wrong
image

package.json
BondarenkoDmitriy: ok

src/makeInfinityAdder.js
jstmpelowycz, line 13: you've got it overdone :harold:

jstmpelowycz (approved):

PR #1210, opened 2023-03-26
added infinite function adder

Esteban4769:


Zad0rnik (approved):
Good job, I probably would have used a different name for the variables, but it`s up to you

SerhiiKirik (approved):
Great 🔥

PR #1211, opened 2023-03-26
Develop

Ch0p3r84:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
-}
+  return function add(...args) {
+    if (args.length === 0) {
+      return 0;
+    }
+
+    let result = args[0];
+
+    return function adder(...newArgs) {
+      if (newArgs.length !== 0) {
```
agniya-i, line 17: pay attention that since we work with only one number in params, you can use check for  empty value, not the length

agniya-i (approved):

PR #1216, opened 2023-03-27
add solution

Pozdnya:


DenysKrysanov (approved):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,33 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  let count = 0;
```
jstmpelowycz, line 10: why do you need counter?

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,33 @@
...
+  const adder = num => {
+    if (num === undefined) {
+      count++;
+
+      if (count > 2) {
+        sum = 0;
+        count = 0;
+
+        return sum;
+      }
```
jstmpelowycz, line 21: > why do you need counter?

seem to be redundant

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,33 @@
...
+    }
+
+    sum += num;
+
+    return adder;
+  };
+
+  if (adder() === undefined) {
+    return 0;
+  }
```
jstmpelowycz, line 33: seem to be hardcoded

jstmpelowycz (changes requested):
you can avoid using counter, just check if num is undefined and:
- if so — return current total but don't forget to annullate `sum` (required)
- otherwise — keep on adding

PR #1218, opened 2023-03-27
solve makeInfinityAdder

kukla1989:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,30 @@
...
+      sum = 0;
+
+      return sum;
+    }
+
+    if (num === undefined) {
+      itsFirstCall = true;
+
+      return sum;
+    }
```
jstmpelowycz, line 24: generally, you don't have to check for the first call, just preserve current total, return it, and don't forget to annulate `sum`

jstmpelowycz (changes requested):

SerhiiKirik (approved):
Great 🔥

PR #1240, opened 2023-04-01
inifinite-sum-closures

OlegRak86:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,32 @@
...
+    if (x !== undefined) {
+      sum += x;
+
+      return add;
+    }
+    count++;
+    sum = count > 1 ? 0 : sum;
+
+    return sum;
+  };
```
nazarmatsevych, line 22: We can return adder function inside itself, no need to create a separate function add

nazarmatsevych (changes requested):

vlad-demchuk (approved):

PR #1242, opened 2023-04-01
solved

Deniso4ka23:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,24 @@
...
+  const adder = (...args) => {
+    if (!args.length) {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
+
+    const [value] = args;
```
polosanya, line 19: Now this line is redundant

polosanya (changes requested):
Almost done! Let's improve your solution a bit)

agniya-i (approved):

PR #1257, opened 2023-04-10
Solution

MaksymKos:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  const result = (x) => {
```
etojeDenys, line 10: Consider using a more descriptive name for the `result` function. For example, you could use `adder` to reflect its purpose.

etojeDenys (approved):
lgtm

PR #1260, opened 2023-04-11
Solution

Dmytro-Kretov:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(num) {
+    if (num !== undefined) {
```
OleziO, line 12: Replace with `num`, it will be the same as `num !== undefined`

OleziO (changes requested):

PR #1261, opened 2023-04-12
Solution

kasiakubisiak:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(addedNumber) {
+    if (addedNumber !== undefined) {
+      sum += addedNumber;
+
+      return adder;
+    } else {
```
solaryasha, line 16: Do you need `else` here ? `if` code returns some value, so you don't need to use `else`.

```suggestion
  if (condition) {
    return value;
  }
  
  // your code
```

solaryasha (changes requested):

solaryasha (approved):

PR #1268, opened 2023-04-18
solution

mariuszgit:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+    if (number !== undefined) {
+      result += number;
+
+      return adder;
+    } else {
+      const tempResult = result;
+
+      result = 0;
+
+      return tempResult;
```
DorotaLeniecDev, line 21:
```suggestion
    if (number !== undefined) {
      result += number;

      return adder;
    } 
      const tempResult = result;

      result = 0;

      return tempResult;
```

DorotaLeniecDev (approved):
Very good, remove else - it will make code more readible

PR #1271, opened 2023-04-19
Solution

oksana-skuba:


src/makeInfinityAdder.js
```diff
@@ -5,8 +5,23 @@
  * @return {function}
  */
 
-function makeInfinityAdder() {
-  // write code here
+function makeInfinityAdder(x) {
+  let result = 0;
```
marina-bidenko, line 9: It is better to name the variable more clearly - for example, sum.

marina-bidenko (approved):
Good job, check minor comment, please

PR #1273, opened 2023-04-20
Solution

antonina-klishch:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(x) {
+    if (x === undefined) {
+      const getSum = sum;
```
YegorKorenkov, line 13: It is better to name the variable more clearly, if a verb is used in the name of the variable, then this is more suitable for the name of the function, since the function performs some action. For the names of variables that store some number, string, etc. better to use a noun, for example _total_

YegorKorenkov (approved):
Well done, please pay attention on comment

PR #1274, opened 2023-04-20
solution

NikaNika12:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result = 0;
+
+  const adder = (a = null) => {
+    if (a === null) {
```
```suggestion
    if (a === undefined) {
```
You can just check if it's undefined in case parameter wasn't passed

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      const newResult = result;
+
+      result = 0;
+
+      return newResult;
+    } else {
+      result += a;
```
polosanya, line 19:
```suggestion
      return newResult;
    }
    
    result += a;
```
No need to add `else` here, as your code will not go further after `return` anyway

polosanya (changes requested):
Almost perfect!

ivanchuk-yurii (approved):

PR #1277, opened 2023-04-24
Develop

AnastasiiaVorobets:


src/makeInfinityAdder.js
```diff
@@ -1,12 +1,23 @@
...
+      sum += number;
+
+      return adder;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
VitaliyBondarenko1982, line 17:
```suggestion
    } 
    const result = sum;

    sum = 0;

    return result;
```

try to avoid else where it is possible. it will increase readibility of code

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #1283, opened 2023-04-26
Solution

Ir-ra:


qoa11a (changes requested):
See comments.

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (number === undefined) {
```
OleziO, line 12: Replace `number === undefined` with` !number`

OleziO (approved):

PR #1284, opened 2023-04-26
solution / js_task-inifinite-sum-closures

stnayn:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
+      sum += num;
+
+      return adder;
+    }
```
tetiana-ivanchenko, line 22:
```suggestion
    if (num === undefined) {
      const result = sum;

      sum = 0;

      return result;
    }
    sum += num;

    return adder;
```

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1285, opened 2023-04-28
Solution

ollavka:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (...args) => {
```
tetiana-ivanchenko, line 11: here should be just one argument

tetiana-ivanchenko (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (num = null) => {
+    if (num === null) {
```
anstsot, line 12: you can use just the check for `undefined` without setting the default value. but your solution is good as well

anstsot (approved):

PR #1297, opened 2023-05-05
solution

luka4ok:
Please, check my solution.

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (number === undefined) {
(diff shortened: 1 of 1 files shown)


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (number === undefined) {
```
OleziO, line 12: Replace `number === undefined` with `!number`

OleziO (approved):

src/makeInfinityAdder.js
luka4ok: But it will cause errors if someone put 0 (zero) to the chain. e.g. adder(4)(0)( ). Doesn't it?

PR #1298, opened 2023-05-06

Donovanoff:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      sum += num;
+
+      return adder;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
TarasHoliuk, line 22:
```suggestion
    if (num !== undefined) {
      sum += num;

      return adder;
    } 
    
    const result = sum;

    sum = 0;

    return result;
```
`else` block is redundant when you return inside `if` block - FYI

TarasHoliuk (approved):

PR #1300, opened 2023-05-08
Solution

NVlasenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(y) {
+    if (y === undefined) {
```
OleziO, line 12: Replace `y === undefined` with `!y`, it will be the same.

OleziO (changes requested):

qoa11a (approved):

PR #1301, opened 2023-05-08
Solution

NadiiaR:


src/makeInfinityAdder.js
```diff
@@ -1,12 +1,35 @@
 'use strict';
 
 /***
+ * Коли в назві завдання присутне слово пекло, нічого хорошого чукати нше варто.
```
OleziO, line 4: Remove all comments.

OleziO (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+  function adder(...nums) {
+    if (nums.length === 0) {
+      const resultSum = sum;
+
+      sum = 0;
+
+      return resultSum;
+    }
+
+    sum += nums.reduce((total, elem) => total + elem);
```
etojeDenys, line 20: The initial value of 0 should be provided as the second parameter to the 'reduce' function to handle cases when the 'nums' array is empty.

etojeDenys (approved):
lgtm

PR #1317, opened 2023-05-17
Adder ready!

vanya-kalyenichenko:


src/makeInfinityAdder.js
```diff
@@ -4,9 +4,18 @@
  *
  * @return {function}
  */
-
 function makeInfinityAdder() {
-  // write code here
+  const adder = function(a) {
+    if (a === undefined) {
+      return 0;
```
anstsot, line 10: don't forget to save previous sum because it shows 0 in case:
```
adder(5)(5)(5)
adder(4)
adder()
```
but it should be 19

anstsot (changes requested):
It’s almost done! Let’s improve a little

SanyaBratashchuk (approved):

PR #1319, opened 2023-05-18
added function makeInfinityAdder

AlexKherson:


src/makeInfinityAdder.js
```diff
@@ -5,8 +5,24 @@
  * @return {function}
  */
 
-function makeInfinityAdder() {
-  // write code here
+function makeInfinityAdder(initialValue = 0) {
```
33Orange, line 8: Consider removing the default value for the 'initialValue' parameter, as it is not mentioned in the task description.

src/makeInfinityAdder.js
```diff
@@ -5,8 +5,24 @@
...
+  let total = initialValue;
+
+  const fn = curr => {
+    if (curr === undefined) {
+      const result = total;
+
+      total = 0;
```
33Orange, line 15: Avoid modifying the total back to 0 after returning the result. It should retain its value for future additions.

src/makeInfinityAdder.js
```diff
@@ -5,8 +5,24 @@
  * @return {function}
  */
 
-function makeInfinityAdder() {
-  // write code here
+function makeInfinityAdder(initialValue = 0) {
+  let total = initialValue;
+
+  const fn = curr => {
```
33Orange, line 11: `const fn ` use more descriptive names

33Orange (changes requested):
Good job.

A few moments with readability

src/makeInfinityAdder.js
silchyn: @AlexKherson ?

silchyn (changes requested):

silchyn (approved):

PR #1324, opened 2023-05-20
HellishChallenge

Dmytro-Tymchenko:


src/makeInfinityAdder.js
```diff
@@ -1,12 +1,23 @@
...
+      sum += num;
+
+      return adder;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
VitaliyBondarenko1982, line 17:
```suggestion
    } 
    
    const result = sum;

    sum = 0;

    return result;
```
avoid else statement where it possible

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #1329, opened 2023-05-22
solution

artem-ivashchenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
...
+        if (y !== undefined) {
+          summary += x + y;
+
+          return adder(x + y);
+        }
+        summary = 0;
+
+        return x;
+      };
+    }
```
polosanya, line 23:
```suggestion
    if (x !== undefined) {
      summary += x;

      return adder;
    }
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
...
+
+    return summary;
```
polosanya, line 25: And if `x === undefined`, you should return the final sum and reset summary variable

polosanya (changes requested):
Good job! Let's optimize your solution a bit)

src/makeInfinityAdder.js
artem-ivashchenko: So easy that way XD? I have been implementing this task for 5 hours straight through two inner functions XD

anstsot (approved):
Good job, approved

PR #1332, opened 2023-05-23
Add task solution

IrynaBalandiukh:


dinaminahi (changes requested):
Something is wrong here

how it supposed to work: 
```
const adder = makeAdder();
adder() === 0
adder(4)(5)() === 9
adder() === 0
adder(5)(5)(5)
adder(4)
adder() === 19
adder() === 0
```

and how it works: 

```
const adder = makeAdder();
adder() === 0
adder(4)(5)() === 9
adder() === 0
adder(5)(5)(5)
adder(4)
adder() !== 19   // here is a problem
adder() === 0
```

image

PR #1336, opened 2023-05-23
add solution

niaYaro:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,28 @@
...
+
+  return function adder(a, b = adderResult) {
+    counter++;
+
+    if (!a && counter === 1) {
+      adderResult = 0;
+      return adderResult
+    }
+
+    if (a) {
```
dinaminahi, line 20: Test is failing when 0 is passed, but instead the 0 should be added, and you should return function until `a` is undefined

dinaminahi (changes requested):

PR #1347, opened 2023-05-26
Solution

yuliia-nudyk:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let total = 0;
+
+  const add = (...number) => {
```
qoa11a, line 11:
```suggestion
  const add = (...args) => {
```

qoa11a, line 11: And it'd be even better if you just pass an argument here instead of using the rest operator. You could check if it's undefined later.

qoa11a (changes requested):

Volodymyr-Melnyk-Fron10der (approved):

PR #1348, opened 2023-05-26
Solution

denys2:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(num) {
+    if (num !== undefined) {
+      sum += num;
+
+      return adder;
+    } else {
```
Anastasiia-Svintsova, line 16: No need for else here. The function will stop if the first condition will be true, because of the return at the end

Anastasiia-Svintsova (approved):

PR #1367, opened 2023-06-13
Solution

reseneweb:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,34 @@
...
-  // write code here
+  let sum = 0;
+  let result = 0;
+
+  function adder(a) {
+    if (a === undefined) {
+      sum += 0;
+      result += 0;
+    } else {
+      sum += a;
```
witflash, line 17: You can use such syntax:
```
sum += (a || 0);
result += (a || 0);
```
to avoid using argument checking.

witflash (approved):
Good job!

PR #1370, opened 2023-06-16
solve the task

eugene-kulish:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const addFunction = function(number) {
+    if (number !== undefined) {
```
OleziO, line 12: Replace with a `if(number)`, it will be the same.

OleziO (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const addFunction = function(number) {
+    if (number) {
```
VitaliyBondarenko1982, line 12: looks like prev solution was better.
this check not cover case when `number === 0`
it will work wiht some follow checks
`number !== undefined`
`number || number === 0`
`typeof number === 'number'`

VitaliyBondarenko1982 (approved):
Well done

PR #1375, opened 2023-06-18
add solution

akozlovska:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,17 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  return function sum(a) {
+    if (arguments.length === 0) {
+      return 0;
```
TarasHoliuk, line 11: this will not work (from example):
```
adder(5)(5)(5)
adder(4)
adder() === 19
```
you should store the current sum between calls and reset it when it's needed instead of immediately return 0

TarasHoliuk (changes requested):

demetrxx (approved):
Well done 👍 Go ahead!

PR #1376, opened 2023-06-18
add task solution

KandaskalovOleg:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+  let sum = 0;
+
+  function adder(num) {
+    if (num === undefined) {
+      const currentSum = sum;
+
+      sum = 0;
+
+      return currentSum;
+    } else {
```
GGsmou, line 18: It wasn't necessary to use `else` here

GGsmou (approved):
GJ!
Approving, but check some recommendations)

src/makeInfinityAdder.js
tetiana-ivanchenko, line 18: remove else

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1379, opened 2023-06-18
solution

vlad-fil:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+  return function adler(num) {
+    if (num === undefined) {
+      const add = sum;
+
+      sum = 0;
+
+      return add;
+    };
+
+    sum += num;
```
mykhalenych, line 20: Avoid using a for loop to calculate the sum of an array, use 'reduce' instead. Replace this line to sum the arguments using 'reduce'.

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adler(num) {
```
mykhalenych, line 11: The adder function should accept multiple arguments, consider using a rest parameter (...args).

mykhalenych (changes requested):

VitaliyBondarenko1982 (approved):

PR #1384, opened 2023-06-20
Solution

adv555:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(num) {
+    if (!num && num !== 0) {
```
maxyma-sanx, line 12: We can also replace with num === undefined

maxyma-sanx (approved):
Good work!

tetiana-ivanchenko (approved):

PR #1388, opened 2023-06-21
add task solution

kbekher:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,33 @@
...
+    };
+
+    if (a) {
+      result += a;
+
+      return sum;
+    }
+
+    return result;
+  };
```
IvanFesenko, line 33: let simplify a bit
Check if there is an argument - if so - we add to result, if not we reset result and return value

```suggestion
  const adder = (a) => {
    if (a === undefined) {
      const total = result;
      result = 0;
      
      return total;
    }
    
    result += a;
    
    return adder;
  };
```

IvanFesenko (changes requested):

mykhalenych (approved):

PR #1389, opened 2023-06-21
infinityAdder solution

ivanbilotserkivskiy:


IraShtynda (approved):
Good job!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,24 @@
...
-  // write code here
+  let sum = 0;
+  let emptyCall = 0;
+
+  const adder = (num) => {
+    if (typeof num === 'undefined') {
+      sum = emptyCall > 0
+        ? 0
+        : sum;
+      emptyCall++;
```
vladmakohon, line 17: It looks like an unnecessary complication in this situation. Furthermore, even though it may pass the tests, it does not work as expected. To illustrate, try testing your function in the following scenario:
``` js
const myAdder = makeInfinityAdder();

myAdder(5)(5)(5);
myAdder(4);
console.log(myAdder()); // Expected: 19. Received: 19
myAdder(5)(5)(5);
myAdder(4);
console.log(myAdder()); // Expected: 19. Received: 0
```
Please ensure that you carefully read the task. Once you verify that the function is called without any parameters, you should clear the sum variable and return the cleared value (using a temporary variable may be helpful here).

vladmakohon (changes requested):
Good job on solving this task! However, I found some issues that require your attention🙂

vladmakohon (approved):
Good work!

tetiana-ivanchenko (approved):

PR #1392, opened 2023-06-21
Solution

dianadiv:


AlyonaSarapina (approved):
nice solution!

tetiana-ivanchenko (changes requested):
this case will not work for you
image

anstsot (approved):

PR #1395, opened 2023-06-21
add task solution

AlyonaSarapina:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
...
+
+    while (num !== undefined) {
+      arr.push(num);
+      num++;
+
+      return getValue;
+    }
+
+    const sum = arr.reduce((a, b) => a + b, 0);
+
```
tetiana-ivanchenko, line 22: use number instead of pushing into array and calculate sum

tetiana-ivanchenko (changes requested):

YegorKorenkov (approved):
Good job!

PR #1397, opened 2023-06-22
add task solution

parmenid89:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result = 0;
+
+  return function sum(a) {
+    if (arguments.length === 0) {
```
anstsot, line 12: it would be better to check `a` variable directly

anstsot (approved):
But pay attention to my comment to avoid the same mistakes in the future

src/makeInfinityAdder.js
parmenid89: ok, tnx👍

PR #1401, opened 2023-06-22
solution

MariiaMonakhova:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = function(...args) {
```
anstsot, line 11: you will always get only one param

anstsot (approved):

PR #1402, opened 2023-06-22
Solution

valet2102:
git commit -m Solution

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  const adder = (arg) => {
+    if (arg === undefined) {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
+
+    sum += arg;
+
+    return adder;
+  };
+
+  return adder;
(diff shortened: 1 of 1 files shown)


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  const adder = (arg) => {
+    if (arg === undefined) {
```
davinnchii, line 13: I think it is better to check if func parameter is number, cuz in tests we don't have a case where we try to pass any other type than number. So with this our function will be better IMHO.
Ofc you will need to change ur function a bit.

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  const adder = (arg) => {
```
davinnchii, line 12: It is better to change `arg` to something like `num`/`number`/`n`, something similar to number or math)

davinnchii (changes requested):
Good job, but I suggest some changes to improve your code!

PR #1409, opened 2023-06-28
Infinity adder function

ValeraViachalo:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
+      sum += num;
+
+      return adder;
+    }
```
```suggestion
    } 
      
    sum += num;

    return adder;

```

redundant `else`

IvanFesenko (approved):

PR #1412, opened 2023-06-28
Solution

dmytro-popovych:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+  return function adder(arg) {
+    if (arg === undefined) {
+      const total = sum;
+
+      sum = 0;
+
+      return total;
+    }
+
+    sum += arg;
```
mykhalenych, line 20: The sum of the arguments is being calculated using a simple addition operator. This should be replaced with the `reduce` function for calculating the sum according to the task requirements.

mykhalenych (approved):

PR #1414, opened 2023-07-02
Solution

katiailchenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  const adder = (firstValue) => {
```
ivadimko, line 10: Consider changing the parameter name 'firstValue' to something more descriptive like 'args' or 'value'

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+    if (!Number.isInteger(firstValue)) {
```
ivadimko, line 11: great move
but smth like `typeof firstValue === 'undefined'` will be more stable

ivadimko (approved):

PR #1421, opened 2023-07-08
Solution

Luk2asz:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,101 @@
  */
 
 function makeInfinityAdder() {
```
DorotaLeniecDev, line 8: In this task the easiest way to create InfiniteAdder would be to create function factory and return the function which can return itself with use of Recursion https://developer.mozilla.org/en-US/docs/Glossary/Recursion.

Try to modify your code using this example as a reference:
```js
function recurse() {
    if(condition) {
        // stop calling itself
        //...
    } else {
        recurse();
    }
}
```

DorotaLeniecDev (changes requested):
Try recursion to avoid duplications

src/makeInfinityAdder.js
Luk2asz: I try with code like below but it is not working

 ` let result = 0;

  function sum(a) {
    if (a === undefined) {
      return result;
    } else {
      result += a;
      sum();
    }
  }

  return sum;`

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      sum += a;
+
+      return add;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
DorotaLeniecDev, line 22: When you are using return in if statement you could remove else statement
```suggestion
    if (a !== undefined) {
      sum += a;

      return add;
    } 
      const result = sum;

      sum = 0;

      return result;
    
```

DorotaLeniecDev (approved):
Good job :) I see you manage to find a solution 💪 , one comment from my side but I will approve the task

PR #1427, opened 2023-07-11
Solution

AllaSerhiienko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,33 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (firstNumber) => {
+    if (firstNumber === undefined) {
```
loralevitska, line 12:
```suggestion
    if (!firstNumber) {
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,33 @@
...
+      return sum;
+    }
+
+    sum += firstNumber;
+
+    const nextAdder = (nextNumber) => {
+      if (nextNumber === undefined) {
```
loralevitska, line 19:
```suggestion
      if (!nextNumber) {
```

loralevitska (changes requested):
Almost done) Let's improve

src/makeInfinityAdder.js
AllaSerhiienko: But if I use !firstNumber, then the number 0 will be false, right?

src/makeInfinityAdder.js
AllaSerhiienko: But if I use !nextNumber, the number 0 will be false, right?

src/makeInfinityAdder.js
VitaliyBodnar: you're right, you can skip this comment

src/makeInfinityAdder.js
VitaliyBodnar: you're right, you can skip this comment

VitaliyBodnar (approved):
Good job!

PR #1428, opened 2023-07-12
Solution

dmytrobetsan:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+
+      return result;
+    } else {
+      sum += num;
+
+      return adder;
+    }
+  }
+
+  return adder;
```
Andriy-Shymkiv, line 25: you can just return an arrow function like this if you want:
```js
  let sum = 0;

  return (num) => {
    if (num === undefined) {
      const result = sum;
      sum = 0;

      return result;
    } else {
      sum += num;

      return adder;
    }
  }
}

```

Andriy-Shymkiv (approved):
good job!

PR #1433, opened 2023-07-15

KarolinaWiktoria:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
+      sum += num;
+
+      return adder;
+    }
```
DorotaLeniecDev, line 22: No need to use **else** here:
```suggestion
    if (num === undefined) {
      const result = sum;

      sum = 0;

      return result;
    } 
      sum += num;

      return adder;
    
```

DorotaLeniecDev (changes requested):
Remove else statement and we are good to go!

src/makeInfinityAdder.js
KarolinaWiktoria: okay :). Already fixed!

DorotaLeniecDev (approved):
Great 👌

PR #1435, opened 2023-07-17
Solution

Illia-L:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = function(num) {
+    if (typeof num !== 'number') {
```
Viktor-Kost, line 12: just check for undefined 
```suggestion
    if (num === undefined) {
```

Viktor-Kost (approved):
Nice!
Just take a look at recommendation I left

PR #1440, opened 2023-07-18
Added task solution

ruslan-ihnatenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+    if (b !== undefined) {
+      result += b;
+
+      return sum;
+    } else {
+      temp = result;
+      result = 0;
+
+      return temp;
+    }
```
loralevitska, line 23: after the special return word, the code is no longer executed, but exits the function, so there is no need to write else
```suggestion
    if (b !== undefined) {
      result += b;

      return sum;
    }
    
    temp = result;
    result = 0;

    return temp;
```

loralevitska (approved):
Well done)

PR #1441, opened 2023-07-19
Solution

Anastasiia-Shahoval:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(num) {
+    if (num === undefined) {
```
xiimiik, line 12: Checking through `undefined` is not reliable, let's use the `typeof`

xiimiik (changes requested):
GJ, almost done!

denys-danyliuk (approved):

PR #1448, opened 2023-07-24
make Infinity Adder

NikolaienkoN:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  const adder = (num) => {
```
loralevitska, line 10: Don't cut variable name
```suggestion
  const adder = (number) => {
```

TarasHoliuk (approved):

PR #1449, opened 2023-07-24
Solution 24072023

Dmytro-Budynskyi:


VadimMentorQA (approved):
Do not include package.json and package-lock.json files into PR

PR #1458, opened 2023-07-28
Solution

mukutiuk:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      sum += arg;
+
+      return (a) => asas(a);
+    } else {
+      const tmp = sum;
+
+      sum = 0;
+
+      return tmp;
+    }
```
```suggestion
    } 
    
    const tmp = sum;

    sum = 0;

    return tmp;
  
```
try to avoid else where it is possible

VitaliyBondarenko1982 (approved):

PR #1462, opened 2023-07-31
Add solution

ArtemBehter:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      result += number;
+
+      return adder;
+    } else {
+      const tempResult = result;
+
+      result = 0;
+
+      return tempResult;
+    }
```
DorotaLeniecDev, line 22: If you are using return statements in if, you could simplify like so:
```suggestion
    if (number !== undefined) {
      result += number;

      return adder;
    } 
      const tempResult = result;
      result = 0;

      return tempResult;
    
```

DorotaLeniecDev (changes requested):
Good job, one suggestion from my side

SanyaBratashchuk (approved):

PR #1467, opened 2023-08-01
Just a solution to infinity adder :3

AndrzejPec:


src/makeInfinityAdder.js
```diff
@@ -1,12 +1,25 @@
...
- *
  * @return {function}
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(value) {
+    if (value === undefined) {
```
choeqq, line 11: I think !value would be a better check - we would handle more falsy cases

choeqq (approved):
Nicely done, please take a look at my comment and you're good to go ✅

src/makeInfinityAdder.js
AndrzejPec: In fact !value lets 0 pass as value "value" inside the if block, which is not desired - one of the tests fail:
 _Should work with chains containing zero as argument (1ms)_

PR #1472, opened 2023-08-04
Develop

Matylda-Lewandowska:
Task Solution -hellish-smth -> task infinite sum closures

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -463,13 +463,248 @@
       }
     },
     "@mate-academy/scripts": {
-      "version": "0.3.5",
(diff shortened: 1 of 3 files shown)


JoyCoffeeAddict (changes requested):
You actually don't need to store memory as an array, you should only care about current sum. That is important, because array operations are `more expensive` than normal arithmetic operations. Tag me if you need any help.

Matylda-Lewandowska:
I hope now it's better! :) I changed memory from array to number (sum of numbers)

choeqq (approved):

PR #1473, opened 2023-08-04
Solution

klaudia-klis:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,31 @@
...
-  // write code here
+  let result;
+
+  function sum(x) {
+    function add(y) {
+      if (y === undefined) {
+        result = 0;
+
+        return x;
+      } else {
```
choeqq, line 17: you can omit else when, you're returning both in if statement and else

choeqq (changes requested):

choeqq (approved):

PR #1474, opened 2023-08-05
init commit

domalewski:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+  let sum = 0;
+
+  return function adder(args) {
+    if (args === undefined) {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
```
choeqq, line 18: when we're returning both in if and else you can omit else statement, it will be reached eventually 😄

choeqq (changes requested):

PR #1476, opened 2023-08-05
First commit infinite adder

hyrych-max:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,27 @@
...
 
 function makeInfinityAdder() {
-  // write code here
+  let sumParametr = 0;
+
+  const result = function(num) {
+    if (num === undefined) {
+      if (sumParametr === 0) {
+        return 0;
+      }
```
maxim2310, line 15: its redundant, code will work same without this

maxim2310 (changes requested):
Good job!!!
just refactor a bit

maxim2310 (approved):
Good job!!!

PR #1480, opened 2023-08-07
Solution

Reject-4444:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result = 0;
+
+  const adder = (num) => {
+    if (num === undefined) {
```
xiimiik, line 12: what if there is a string or null?  try to use `typeof`

xiimiik (changes requested):

Andriy-Shymkiv (approved):

PR #1485, opened 2023-08-09
solution

NataliMax:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result = 0;
+  let currentResult = 0;
+
+  const adder = (a) => {
+    if (a > -1) {
```
IvanFesenko, line 13:
```suggestion
    if (a !== undefined) {
```

IvanFesenko (approved):
Pay attention on my comment

src/makeInfinityAdder.js
NataliMax: Thanks!

PR #1486, opened 2023-08-09
made infinity adder

ivan-baranovskyi:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum;
+  let allNum = [];
+
+  function add(number) {
+    if (number !== undefined) {
+      allNum.push(number);
```
TarasHoliuk, line 14: you don't need `allNum` array - just change `sum` here

TarasHoliuk (changes requested):

TarasHoliuk (approved):
:+1:

PR #1488, opened 2023-08-10
add task solution

voievodik:


Wesses (approved):
Well done!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(num) {
+    if (num === undefined) {
+      const result = sum;
```
alexander-ignatow, line 13: Why are you copying the value here?

alexander-ignatow (changes requested):
Good 👍 

Just one minor thing to fix

src/makeInfinityAdder.js
voievodik: @alexander-ignatow I need to clear value for empty argument call, so I keep this value in new variables, then clear sum and return result;

Moroz-Dmytro (approved):
GJ 👍

PR #1489, opened 2023-08-10
add task solution

Wesses:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (...value) => {
+    if (value.length) {
+      sum += value[0];
+
+      return adder;
+    }
```
33Orange, line 16: Is it necessary to rest params?
Your function should take 1 argument.

```suggestion
  const adder = (number) => {
    if (some cond) {
      sum += number;

      return adder;
    }
```

33Orange (changes requested):
Almost done

AntonHerasymov (approved):
Good job!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (value) => {
+    if (value || value === 0) {
```
```suggestion
    if (value !== undefined) {
```
You can just check if the value is not undefined

polosanya (approved):
Nice! I'm approving it, but please pay attention to the comment)

PR #1491, opened 2023-08-13
feat: add task solution

FrontKid:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
...
+      const sum = collection;
+
+      collection = 0;
+
+      return sum;
+    }
+
+    // If the function takes more than one parameter
+    // sum it with reduce
+    collection += nums.reduce((acc, n) => acc + n, 0);
```
desevoker, line 22:
```suggestion
  let sum = 0;

  const adder = (...nums) => {
    if (!nums.length) {
      const result = sum;

      sum = 0;

      return result;
    }

    // If the function takes more than one parameter
    // sum it with reduce
    sum += nums.reduce((acc, n) => acc + n, 0);
```
Consider better variable names.

desevoker, line 22: Reduce looks like overkill.

desevoker (approved):
Good job! 🚀

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let collection = 0;
+
+  const adder = (...nums) => {
```
33Orange, line 11: Adder function should take only 1 argument.

image

```suggestion
  const adder = (number) => {
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
...
+    if (!nums.length) {
```
33Orange, line 12:
```suggestion
    if (number !== undefined) {
```

src/makeInfinityAdder.js
33Orange, line 22:
```suggestion
    collection += number;
```

33Orange (changes requested):
Almost done

Moroz-Dmytro (approved):
👍

PR #1492, opened 2023-08-13
Solution

desevoker:


src/makeInfinityAdder.js
```diff
@@ -1,12 +1,27 @@
...
- *
  * @return {function}
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (typeof number !== 'number') {
```
AndriiKuz, line 11: Maybe better use arg === undefined? insted `typeof number !== 'number'`

AndriiKuz (changes requested):

src/makeInfinityAdder.js
desevoker: Can you explain why it may be better?
I don't see a point why we should allow arguments of some other type than number. How should the function handle them? There are no requirements or tests for that, only for numbers.

tiserett (approved):

PR #1498, opened 2023-08-14
initial commit

maksym-mishchanchuk:
I have a problem with tests. The mentor checked on qna, but did not find an error. Made a push with the no verify flag

locally, the tests all run and console

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(number) {
+    if (typeof number !== 'number') {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
+
+    sum += number;
+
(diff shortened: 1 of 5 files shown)


yuliiaant (approved):
Looks great!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(number) {
+    if (typeof number !== 'number') {
```
tiserett, line 12: better actually check if it's undefined

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
+
+    sum += number;
```
tiserett, line 20: and here you can check if it's number (if you want to)

tiserett (approved):

PR #1501, opened 2023-08-14
add task solution

AngerDESTROYS:


00Mass00 (approved):
Good job!🚀

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,26 @@
 
 function makeInfinityAdder() {
   // write code here
+  let SUM = 0;
+  let CURRENT_NUMBERS = [];
```
Moroz-Dmytro, line 11: Variables should be named in camelCase

Moroz-Dmytro, line 11: Why do you need this variable? It is useless

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,26 @@
...
+
+  function adder(...args) {
```
Moroz-Dmytro, line 13: Function have only one argument, you don't need rest here

Moroz-Dmytro (changes requested):
Let's improve your solution

roman-mirzoian (approved):
Well done 👍

PR #1503, opened 2023-08-15
Solution

yuliiaant:


KatOlista (approved):
good job!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let counter = 0;
```
Moroz-Dmytro, line 9: But I'd rather call it `total` or `sum`

Moroz-Dmytro (approved):
👍

PR #1504, opened 2023-08-15
Solution

Oleh0293:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,24 @@
 
 function makeInfinityAdder() {
   // write code here
-}
+  let sum = 0;
+  let currentArgs = [];
+
+  function adder(...args) {
+    if (args.length === 0) {
+      const result = sum + currentArgs.reduce((acc, val) => acc + val, 0);
```
polosanya, line 15:
```suggestion
  let sum = 0;

  function adder(number) {
    if (number === undefined) {
      const result = sum;
```
currentArgs is actually redundant. You can just use recursion with one parameter instead

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,24 @@
...
+
+      sum = 0;
+      currentArgs = [];
 
+      return result;
+    } else {
+      currentArgs = [...currentArgs, ...args];
+
+      return adder;
```
polosanya, line 24:
```suggestion
      return result;
    }
      
    sum += number;

    return adder;
```
`else` is also redundant, as your code will not go further after return anyway

polosanya (changes requested):
Good job! Let's try to improve your solution a bit)

tiserett (approved):

PR #1512, opened 2023-08-16
solution

Taras2705:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      totalSum += num;
+
+      return adder;
+    } else {
+      const result = totalSum;
+
+      totalSum = 0;
+
+      return result;
+    }
```
Moroz-Dmytro, line 22: Try to improve this. When `if` returns result we don't need `else`

Moroz-Dmytro (changes requested):
GJ 👍 
Small comment

BudnikOleksii (approved):
Good job!

PR #1515, opened 2023-08-18
add task solution

ydem1:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,21 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  return function adder(add = false) {
+    sum += add;
```
Anastasiia-Svintsova, line 13: The function 'adder' could be an arrow function for more concise and readable code.

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,21 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  return function adder(add = false) {
```
Anastasiia-Svintsova, line 12: It's not the best approach to assign the value to false as it could lead to confusion and unexpected behavior. You can check if it's undefined instead

Anastasiia-Svintsova (changes requested):
Almost done! Check my comments

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  const adder = add => {
+    if (add === undefined) {
```
Esceype, line 13:
```suggestion
    if (!add) {
```

Esceype (approved):
GJ!!! But read the comment to make your code looks better in the future

PR #1517, opened 2023-08-19
Solution

Orpheus29:


Valentyn-Bilyk (approved):
Well done!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  const adder = (arg) => {
+    if (arg === undefined) {
```
roman-mirzoian, line 11: It is not necessary to compare exactly with `undefined`, you can simply check whether this argument exists through the logical NOT.
```js
if (!number) {
```

roman-mirzoian (approved):
Well done, but pay attention to the comment 👍

src/makeInfinityAdder.js
Orpheus29: Hello Roman, thank you for the comment. I tried that. It's generally like you say, but in this case the ! version won't pass the test if zero has been received as an argument (1)(1)(1)(1)(0)(1)(1)(1)()

roman-mirzoian: @Orpheus29 yes, you are right 🔥

PR #1519, opened 2023-08-20
solution

Valentyn-Bilyk:


FrontKid (approved):
Good job!!!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let total = 0;
+
+  const adder = function(number) {
+    if (number === undefined) {
```
roman-mirzoian, line 12: It is not necessary to compare exactly with `undefined`, you can simply check whether this argument exists through the logical NOT.

roman-mirzoian (approved):
Well done 👍

PR #1525, opened 2023-08-26
infinity-adder_solution

olha-rypich:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = a => {
```
lerastarynets, line 11: what is `a`? pls mind descriptive naming

lerastarynets, line 11:
```suggestion
  const adder = (num) => {
```

lerastarynets (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

etojeDenys (approved):
lgtm

PR #1528, opened 2023-08-28
Add task solution

spojrzenie:


spojrzenie:
Hi, i made mistake with this PR, because compare is develop -> to develop. Here i made another PR and compare master to my develop: https://github.com/mate-academy/js_task-inifinite-sum-closures/pull/1529

spojrzenie:
> Hi, i made mistake with this PR, because compare is develop -> to develop. Here i made another PR and compare master to my develop: #1529

I get lucky and i made change develop to master in this PR.

carlos-cne (approved):
Your code looks great!
Even though there is another way to solve this algorithm without using the reduce method, like using an inner function that will be called instead of the adder function.

PR #1535, opened 2023-09-01
Solution

GIroSTarK:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (arg) => {
+    if (arg === undefined) {
```
Andriy-Shymkiv, line 12: what if `arg` will be `null`, not `undefined`?

Andriy-Shymkiv (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (arg) => {
+    if (typeof arg !== 'number') {
```
Andriy-Shymkiv, line 12: 🔥

Andriy-Shymkiv (approved):

PR #1536, opened 2023-09-01
infinityCalculator

Kyzenko:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+
+  const calculator = (argument) => {
+    if (typeof argument !== 'number') {
```
ivbutsykin, line 13: I can suggest to make check in such way, but your variant also fine:
```suggestion
    if (!argument) {
```

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,23 @@
 
 function makeInfinityAdder() {
   // write code here
```
ivbutsykin, line 9:
```suggestion

```

ivbutsykin (approved):
Good job!

PR #1537, opened 2023-09-02
solution

GavriliukArtem:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const add = (value) => {
+    if (value === undefined) {
```
Andriy-Shymkiv, line 12: what if value is `null`?

andriy-shymkiv (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #1539, opened 2023-09-04
Solution

ValeriiaFomenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(num) {
```
maxim2310, line 11: you need use function expression in this case

maxim2310 (approved):
Good work!!!

PR #1540, opened 2023-09-06
Solution

HachevskyyMarkiyan:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (value) => {
+    if (!isNaN(value)) {
```
maxim2310, line 12: isNaN can have unexpected behavior and not so understanded
```suggestion
    if (value !== undefined) {
```

maxim2310 (changes requested):
Good job! Let's improve your code a bit)

src/makeInfinityAdder.js
HachevskyyMarkiyan: Thanks!
Code updated✅

maxim2310 (approved):
Good work!!!

PR #1547, opened 2023-09-09
add task solution

Daniil-Lilin:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,27 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  const adder = (num) => {
+    if (num === undefined) {
+      return 0;
+    }
```
TarasHoliuk, line 12: You don't need an additional function (adder) just to handle this case
Just define the initial `sum` as `0`

Note that in this case, you will need to handle resetting sum before returning it

TarasHoliuk (changes requested):

VitaliyBondarenko1982 (approved):
GJ!

PR #1558, opened 2023-09-16
task solution

no4kar:


src/makeInfinityAdder.js
```diff
@@ -1,12 +1,24 @@
...
+      return summer;
+    }
+
+    augend[1] = augend[0];
+    augend[0] = 0;
+
+    return augend[1];
+  };
+
+  return summer;
```
etojeDenys: why do we need array here?

src/makeInfinityAdder.js
```diff
@@ -1,12 +1,24 @@
...
+  const augend = [0, 0];
+
+  const summer = (addend) => {
+    if (addend !== undefined) {
+      augend[0] += addend;
+
+      return summer;
+    }
+
+    augend[1] = augend[0];
```
etojeDenys, line 15: we can just create some var instead of this

etojeDenys (changes requested):

src/makeInfinityAdder.js
no4kar: creation extra variables increase stress to system. more efficiently make buffer.

etojeDenys: I think creating two variables ahead of time will be much better and clearer than one array

etojeDenys (changes requested):
what do you think?

src/makeInfinityAdder.js
no4kar: where is there operation of array's clear? only assignment. buffer will be create one time and used every time for calculation, but now shell needs to create 'tempSum' every time even it won't be used.

andriy-shymkiv (approved):

src/makeInfinityAdder.js
Andriy-Shymkiv: agree with @etojeDenys. Creating an array is unnecessary here

PR #1560, opened 2023-09-17
add task solution

Liubomyr19:
[Guideline](https://Liubomyr19/mate-academy/js_task-guideline/blob/master/README.md)

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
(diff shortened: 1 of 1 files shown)


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      const temp = sum;
+
+      sum = 0;
+
+      return temp;
+    }
+
+    sum += arg;
+
+    return (nextArg) => adder(nextArg);
```
SanyaBratashchuk, line 22:
```suggestion
    return adder;
```

SanyaBratashchuk (approved):

PR #1561, opened 2023-09-18
solution

Patryk-Buczkowski:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      const sum = memory;
+
+      memory = 0;
+
+      return sum;
+    }
+
+    memory += number;
+
+    return (nextNumber) => adder(nextNumber);
```
maxim2310, line 22:
```suggestion
    return adder
```

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

maxim2310 (approved):
Good work!!!

PR #1564, opened 2023-09-19
Solution

InnaSh23:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      const currentSum = sum;
+
+      sum = 0;
+
+      return currentSum;
+    } else {
+      sum += args.reduce((acc, val) => acc + val, 0);
+
+      return adder;
+    }
```
Viktor-Kost, line 22: no needs to use `else` if you use return earlier
and you can simplify avoiding reduce method
```suggestion
    }
    
    sum += number;

    return adder;
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  return function adder(...args) {
```
Viktor-Kost, line 11: just use one value instead of args
```suggestion
  return function adder(number) {
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+    if (args.length === 0) {
```
Viktor-Kost, line 12: compare with undefined
```suggestion
    if (args.length === undefined) {
```

Viktor-Kost (changes requested):
GJ!
Some minor fixes needed

nazarmatsevych (approved):
👍

PR #1566, opened 2023-09-21
v1

DVasylenko18:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,27 @@
...
+
+      return storedSum;
+    }
+
+    return (b) => {
+      if (b === undefined) {
+        return storedSum;
+      } else {
+        return adder(b);
+      }
```
maxim2310, line 26: this redundant because we never go here

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,27 @@
 
 function makeInfinityAdder() {
   // write code here
+  let storedSum = 0;
+
+  function adder(a) {
+    if (a !== undefined) {
+      storedSum += a;
```
maxim2310, line 14:
```suggestion
      storedSum += a;
      return adder
```

src/makeInfinityAdder.js
```diff
@@ -7,6 +7,27 @@
...
+    } else {
+      storedSum = 0;
+
+      return storedSum;
```
maxim2310, line 18:
```suggestion
      const result = storedSum
      storedSum = 0;

      return result ;
```

maxim2310 (changes requested):
Good job! Let's improve your code a bit)

Andriy-Shymkiv (approved):

PR #1570, opened 2023-09-22
add task solution

Vovchukkul:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
...
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
+      sum += num;
+
+      return add;
+    }
```
```suggestion
    if (num === undefined) {
      const result = sum;

      sum = 0;

      return result;
    } 
    
    sum += num;

    return add;
    }
```
redundant else block

etojeDenys (approved):
good job

PR #1572, opened 2023-09-26
`SOLUTION`

sviatikkondur:
Hello, check my work!

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (number !== undefined) {
+      sum += number;
+
+      return adder;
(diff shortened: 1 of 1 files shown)


Helen-arch (approved):
Nice!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+    }
+
+    const result = sum;
+
+    sum = 0;
+
+    return result;
+  };
+
+  return adder;
```
StasSokolov, line 25:
```suggestion
  return adder = (number) => {
    if (number !== undefined) {
      sum += number;

      return adder;
    }

    const result = sum;

    sum = 0;

    return result;
  };
```

StasSokolov (changes requested):
Great job, only one minor comment from me!

StasSokolov (approved):
Great job🔥

PR #1579, opened 2023-09-28
add Hellish challenge task solution

o-drozzdyk:


koros-rk (approved):
I think it can be done without additional function adder() but it still good job ;)

StasSokolov (approved):
Great job, i'm approving it🔥

PR #1580, opened 2023-09-29
Currying adder

ElenaGrasovskaya:


koros-rk (approved):
I think it can be simplified little bit, but it still good job ;)

vasylmiskiv (approved):

PR #1585, opened 2023-10-01
first solution

Clover247:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,21 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+
+  return function func(a) {
```
OlhaKoziuk, line 11: perhaps not a good idea for a name of function, maybe better to call it like sum and we can use function expression instead of function declaration  [7.1. Use named function expressions instead of function declarations.](https://github.com/mate-academy/style-guides/blob/master/javascript.md#71-use-named-function-expressions-instead-of-function-declarations)

superpooperxxx (changes requested):
Good job🔥 Fix the comments of your mate. 

I also slightly don't like the naming:
1. `a` could be `value`
2. `amount` could be `sum`, cause they have different meaning and we calculate sum here.
3. `temporary` is basically a final sum, not something temporary.
4. And off course `func`

StasSokolov (approved):
Awesome job, i'm approving it 🔥🔥

PR #1591, opened 2023-10-03
solution

vlevik:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
-  // write code here
+  let result = 0;
+
+  function adder(num) {
+    if (num !== undefined) {
+      result += num;
+
+      return adder;
+    } else {
+      const finalResult = result;
```
polosanya, line 17:
```suggestion
      return adder;
    }
    
    const finalResult = result;
```
`else` is redundant here as the code will not go further after return anyway

polosanya, line 17: Not critical in this task, but it would be great to use more meaningful names than just `result` and `finalResult`

polosanya (changes requested):
Almost perfect!

ElenaGrasovskaya:
Perfect!

ElenaGrasovskaya (approved):
Perfect!

superpooperxxx (approved):
Good job🔥I agree with Sasha that you should get used to naming your variables more meaningfully

PR #1597, opened 2023-10-08
task solution

TenetaJ:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
+      sum += num;
+
+      return adder;
+    }
```
```suggestion
    } 
    
    sum += num;

    return adder;
```
else here looks like redundant. in any case will be return in if block.

VitaliyBondarenko1982 (approved):
Looks good.

PR #1599, opened 2023-10-09
add solution

Livan94:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
+      sum += num;
+
+      return adder;
+    }
```
VitaliyBodnar, line 22: you can make it less
```suggestion
    if (num === undefined) {
      const result = sum;

      sum = 0;

      return result;
    }
    sum += num;

    return adder;
```

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):
Good job!

PR #1601, opened 2023-10-09
function makeAdder was writed

EkaterinaBessonnaya:
😉

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -6,18 +6,17 @@
  */
 
 function makeInfinityAdder() {
-  let sumAllNumers = 0;
+  let sum = 0;
 
-  const adder = (numbers) => {
-    if (numbers === undefined) {
-      const result = sumAllNumers;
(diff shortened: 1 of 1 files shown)


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sumAllNumers = 0;
```
Andriy-Shymkiv, line 9: better name it just `sum`

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+
+  const adder = (numbers) => {
```
Andriy-Shymkiv, line 11: it's just a single number, because numbers it's like an array or smth
```suggestion
  const adder = (number) => {
```

Andriy-Shymkiv (changes requested):

Viktor-Kost (approved):
GJ!

PR #1606, opened 2023-10-13
Solution

ihor-pavlov:


ihor-pavlov:
I agree with the change of the "if" verification condition. But it is impossible to pick up other lines, because the condition states: "When calling without parameters, it will return the result and clear the amount." Also, the function must return itself so that it can be called again. And even after the changes you specified, the automatic tests do not pass.

anastasiiavorobiova (approved):
Great work!

PR #1608, opened 2023-10-14
Solution

denedweb:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      sum += num;
+
+      return adder;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
```suggestion
    if (num !== undefined) {
      sum += num;

      return adder;
    } 
    
    const result = sum;

    sum = 0;
 
     return result;
    }
```
block else is redundant here

etojeDenys (changes requested):
awesome

sTorba24 (approved):

PR #1610, opened 2023-10-16
Solution

GlibVK:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
 
 function makeInfinityAdder() {
-  // write code here
+  let getSum = 0;
+
+  const adder = (...args) => {
+    if (args.length === 0) {
+      const sum = getSum;
+
+      getSum = 0;
```
VitaliyBodnar, line 15: getSum it sounds a function but it's a variable, so you need to rename it

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let getSum = 0;
+
+  const adder = (...args) => {
```
VitaliyBodnar, line 11: this adder function takes only one argument, so I does not make any sense to use rest operator here

VitaliyBodnar (changes requested):

anastasiiavorobiova (approved):
Awesome work!

PR #1618, opened 2023-10-17
add function make-infinity-adder

AntonBelia:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let arrayResult = [];
```
etojeDenys, line 9: you can rewrite to use result var as a number

etojeDenys: it will simplify the logic

etojeDenys (changes requested):

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,25 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let callArray = [];
```
etojeDenys, line 9: try to rewrite this task to not use arrays at all

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #1620, opened 2023-10-18
Homework: JS Task Infinite Sum Closures

ViktorBykh:


sarakhmen (approved):
Good job! But I'd recommend reducing the number of empty lines as sometimes it affects the code readability

PR #1622, opened 2023-10-19
solution

MariiaSuper:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      const sum = result;
+
+      result = 0;
+
+      return sum;
+    } else {
+      result += number;
+
+      return adder;
+    }
```
VitaliyBodnar, line 22: can be shorter 
```suggestion
    }
    result += number;

    return adder;
```

VitaliyBodnar (changes requested):

VitaliyBodnar (changes requested):
.

lerastarynets (approved):
You're showing incredible skills, keep it going!

PR #1624, opened 2023-10-19
solution

nikkias:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      sum += value;
+
+      return adder;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
```suggestion
    if (value !== undefined) {
      sum += value;

      return adder;
    }
    
    const result = sum;

    sum = 0;

    return result;
```
block else is redundant here

etojeDenys (changes requested):

volodymyr-soltys97 (approved):
Well done 🔥

PR #1631, opened 2023-10-26
Solution Hellish challenge

garmoni:


maxim2310 (changes requested):
Good work, but now you do not store the value and its working if func calls like chain but if i make pause it will not work
image

anastasiiavorobiova (approved):
Great work!

PR #1637, opened 2023-10-29
complete task

Annbpiu:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let char = 0;
+
+  const amout = function(num) {
```
maxim2310, line 11:
```suggestion
  const amount = function(num) {
```

maxim2310 (approved):
Good work!!!

src/makeInfinityAdder.js
Annbpiu: thx for corrections

PR #1657, opened 2023-11-08
solution

yevtok:


src/makeInfinityAdder.js
```diff
@@ -5,8 +5,24 @@
...
-  // write code here
+function makeInfinityAdder(x) {
+  let sum = 0;
+
+  function adder(number) {
+    if (number !== undefined) {
+      sum += number;
+
+      return adder;
+    } else {
```
polosanya, line 16:
```suggestion
      return adder;
    }
```
Else is redundant here. Your code will not go further after `return` anyway

src/makeInfinityAdder.js
```diff
@@ -5,8 +5,24 @@
  * @return {function}
  */
 
-function makeInfinityAdder() {
-  // write code here
+function makeInfinityAdder(x) {
```
polosanya, line 8:
```suggestion
function makeInfinityAdder() {
```
This function shouldn't receive any parameters. By the way, you don't even use it.
Pay attention to the example 
image

polosanya (changes requested):
Almost done!

polosanya (approved):
Well done!

PR #1661, opened 2023-11-09
done task Hellish

Anton-Kuchmasov:


MrHunterBoi (approved):
Perfect

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let counter = 0;
+  const adder = (a) => {
+    if (a !== undefined) {
```
ArtemTeslenko, line 11: do not compare with undefined

ArtemTeslenko (approved):
Good job 👍

src/makeInfinityAdder.js
Anton-Kuchmasov: why? it's a bad practice?

ArtemTeslenko: Yes, it is a bad practice, but maybe in this case it is normal, because you can get `0` and it should be a valid value. Let's leave this comparing here 😉 Great solution 🔥

PR #1663, opened 2023-11-10
add task solution

ruslanagnatenko:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (num) => {
+    const resultSum = sum;
+
+    if (num !== undefined) {
```
MrHunterBoi, line 14: You can inverse this condition so as to not use `else`

MrHunterBoi (approved):
Well done

src/makeInfinityAdder.js
BudnikOleksii: > You can inverse this condition so as to not use `else`

Good point, avoid else statement if you can

BudnikOleksii (changes requested):
Looks good 👍 
Just one moment that [Maksym Svoboda](https://github.com/MrHunterBoi) mentioned

artemforsoff (approved):
Good 🤟🏻

PR #1668, opened 2023-11-13
solution

jacekkosiba:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+    if (x !== undefined) {
+      sum += x;
+      return adder;
+    }
+
+    if (x === undefined) {
+      const result = sum;
+      sum = 0;
+      return result;
+    }
```
mvjl000, line 21: Since you are checking  `x !== undefined` above (and returning inside that if statement), you don't another if here. 
```suggestion
      const result = sum;
      sum = 0;
      return result;
```

mvjl000 (changes requested):
Well done, just one thing to improve.

szymanski-marcin (approved):

mbruhler (approved):
Approved

PR #1675, opened 2023-11-15
complete task

boytrend0108:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,33 @@
...
+      if (b === undefined) {
+        const res = sum;
+
+        sum = 0;
+
+        return res;
+      }
+
+      return adder(b);
+    };
```
nazarmatsevych, line 32: The second function looks redundant, please check if we can simplify the solution

nazarmatsevych (changes requested):

etojeDenys (approved):
awesome

PR #1682, opened 2023-11-22
create infinity addition funtion

mixelio:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+    if (a === undefined) {
+      const lastResult = result;
+
+      result = 0;
+
+      return lastResult;
+    } else {
+      result += a;
+    }
+
```
Viktor-Kost, line 21: no needs to add `else` since you've used `return` at `if`
```suggestion
    }
    
    result += a;

```

Viktor-Kost (approved):
GJ!
Just left a comment for you to know

PR #1685, opened 2023-11-23
Add task solution

Laisens:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,117 @@
...
+                    sum += n;
+                  };
+                };
+              };
+            };
+          };
+        };
+      };
+    };
+  };
```
nazarmatsevych, line 118: Looks too complicated, you can run  function inside itself if this condition is false
`a === undefined`

nazarmatsevych (changes requested):

nazarmatsevych (approved):
Awesome 🔥

PR #1691, opened 2023-11-27
Solution

odolga:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,34 @@
...
+  return function makeAdder(value1) {
+    if (!value1) {
+      if (result !== 0) {
+        sum = result;
+        result = 0;
+
+        return sum;
+      }
+
+      return result;
```
Viktor-Kost, line 21: 1. you should check value for undefined. If it's true - reset result
2. if it's not true just add a value to result
3. `adder` function should return itself
```suggestion
    if (value === undefined) {
      const sum = result;
      result = 0;

      return sum;
    }
      
    result += num

    return adder;
```

Viktor-Kost (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #1696, opened 2023-11-28
Solution

julmak:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      sum += number;
+
+      return adder;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
```suggestion
    }
    
    const result = sum;

    sum = 0;

    return result;
   
```
else is redundant here

etojeDenys (approved):
good job

PR #1697, opened 2023-11-28
Solution

LabPetro:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(...args) {
+    if (args.length === 0) {
```
lerastarynets, line 12:
```suggestion
    if (!args.length) {
```

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #1698, opened 2023-11-29
Task solution :) 

BodyaRespect:


etojeDenys (changes requested):
try to simplify this implementation without array, you can just create the sum var and numbers there

etojeDenys (approved):
awesome

PR #1723, opened 2023-12-18
Solution

dshyshenko:


src/makeInfinityAdder.js
```diff
@@ -7,6 +7,22 @@
 
 function makeInfinityAdder() {
   // write code here
+  let sum = 0;
+  const makeAdder = (args) => {
+    if (args || args === 0) {
+      sum += args;
+
+      return makeAdder;
+    } else if (!args) {
```
nazarmatsevych, line 16: Conditions look too complicated and we can change them a bit

if (args === undefined) {
  
} else {
  
}

nazarmatsevych (approved):
Good work, please check the comment below

PR #1724, opened 2023-12-18
it is real HELL)

Stan1slav123:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,79 @@
...
+                    };
+                  };
+                };
+              };
+            };
+          };
+        };
+      };
+    };
+  };
```
nazarmatsevych, line 79: Too complex a solution, and it will stop working if you update the tests a bit
Read about [recursion](https://developer.mozilla.org/en-US/docs/Glossary/Recursion) and try to rewrite the solution

nazarmatsevych (changes requested):
If you have any questions - feel free to ask them in fe_chat

Viktor-Kost (approved):
Much better)

PR #1726, opened 2023-12-19
add task solution

Svimpik:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
+      sum += value;
+
+      return adder;
+    }
```
Viktor-Kost, line 22: else condition is redundant since you've used early return above
```suggestion
    }
    
    sum += value;

    return adder; 
```

Viktor-Kost (approved):
GJ!
Just apply my recommendation

PR #1728, opened 2023-12-21
add task solution

artem8746:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result = 0;
+
+  const addFunc = (number) => {
```
ssaviak, line 11: By the rule of the task - you had to name the function like "adder"

ssaviak (approved):
Good job!

olya-shyian (approved):
Well done!💥

PR #1734, opened 2023-12-21
infinity adder

vpashko:


DanilWeda (approved):
Thanks for the realization. 
Approved!

PR #1764, opened 2024-01-10
add task solution

dima91020:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+  let sum = 0;
+
+  const adder = (a) => {
+    if (a === undefined) {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
```
etojeDenys, line 18:
```suggestion
    } 
```
looks redundant

etojeDenys (approved):
good job

PR #1766, opened 2024-01-11
task done

taraShulak:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+  let sum = 0;
+
+  function makeAdder(num) {
+    if (num || num === 0) {
+      sum += num;
+
+      return makeAdder;
+    }
+
+    const RESULT = sum;
```
etojeDenys, line 18:
```suggestion
    const result = sum;
```
it's not the constant

etojeDenys (approved):
good job

PR #1793, opened 2024-01-29
Solution

sergii-stognii:
Closures: Hellish challenge.

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(num) {
+    if (num !== undefined) {
+      sum += num;
+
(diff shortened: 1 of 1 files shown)


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      sum += num;
+
+      return adder;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
DarkMistyRoom, line 22:
```suggestion
    if (num !== undefined) {
      sum += num;

      return adder;
    } 
      
    const result = sum;

    sum = 0;

    return result;
```

DarkMistyRoom: you can do it without `else` because you have `return` inside the `if`

DarkMistyRoom (approved):
Good job!

PR #1795, opened 2024-01-29
solution

ant-tur:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      sum += num;
+
+      return adder;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
Viktor-Kost, line 22: `else` is redundant since you've used early `return`
```suggestion
    }
    
    const result = sum;

    sum = 0;

    return result;
```

Viktor-Kost (approved):
Nice!
Just apply a recommendation I left for you

PR #1802, opened 2024-02-04
add task solution

Tunik-hub:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      sum += value;
+
+      return adder;
+    } else {
+      const currentSum = sum;
+
+      sum = 0;
+
+      return currentSum;
+    }
```
volodymyr-soltys97, line 22: else is redundant here
```suggestion
    }
    
    const currentSum = sum;

    sum = 0;

    return currentSum;
```

volodymyr-soltys97 (approved):
Good job 👍

PR #1813, opened 2024-02-09
task done

Reaffith:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
...
-}
+  let sum = 0;
+
+  function adder(num) {
+    if (typeof num === 'number') {
+      sum += num;
+
+      return adder;
+    } else {
+      const result = sum;
```
etojeDenys, line 17:
```suggestion
    } 
    
    const result = sum;
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,26 @@
...
+      sum = 0;
+
+      return result;
+    }
+  }
+
+  if (arguments.length > 0) {
+    return adder(...arguments);
+  } else {
+    return adder;
```
etojeDenys, line 28:
```suggestion

return adder;
```
looks like redundant check

etojeDenys (approved):
good job

PR #1814, opened 2024-02-09
Add solution

Vasya564:


nk8bnj (approved):
Good job!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+  const adder = (...number) => {
+    if (!number.length) {
+      const sumCopy = sum;
+
+      sum = 0;
+
+      return sumCopy;
+    } else {
+      sum += number[0];
+    }
```
BudnikOleksii, line 20:
```suggestion
    } 
    
    sum += number[0];
```
You don't need else as you `return` in the `if`

BudnikOleksii (approved):
Good job 👍

PR #1831, opened 2024-02-17
Solution

HolyRaveOS:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      sum += num;
+
+      return adder;
+    } else {
+      const tempSum = sum;
+
+      sum = 0;
+
+      return tempSum;
+    }
```
BudnikOleksii, line 22: You don't need `else` in this case

BudnikOleksii (approved):
👍

PR #1840, opened 2024-02-21
Develop

TideTwo:
Had an issue with this task. I think either the tests or example might need some editing? My issue when working on it was that the readme shows an example case that is not present in the tests. Specifically:
```
adder(5)(5)(5);
adder(4);
adder() === 19;`
```

here `adder` is called multiple times, but the last call in the first 2 chains contains an argument, whereas **none** of the test showcase this, the above code follwowing the test format would look like:
```
adder(5)(5)(5)();
adder(4)();
adder() === 0  (I think it would be 0?);
```

making it hard to understand the intended behaviour in that specific case. I have gone over the issue with a mentor already.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x]
+        node-version: [14.x]
 
     steps:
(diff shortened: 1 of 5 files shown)


TideTwo:
image

choeqq (approved):

PR #1842, opened 2024-02-21
Solution

kovaltar:
Maybe this task has a better solution. If so - please help to improve

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -6,7 +6,44 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let totalSum = 0;
+  const adder = (firstNumber) => {
+    if (!firstNumber) {
+      const sumToDisplay = totalSum;
+
+      totalSum = 0;
+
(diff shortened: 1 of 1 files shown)


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,36 @@
...
+        return sumToDisplay;
+      }
+
+      totalSum += secondNumber;
+
+      return func;
+    }
+
+    return func;
+  };
```
etojeDenys, line 36:
```suggestion
  function adder(num) {
    if (num === undefined) {
      const result = totalSum;

      totalSum = 0;

      return result;
    }

    totalSum += num;

    return adder;
  }
```

etojeDenys (changes requested):

src/makeInfinityAdder.js
kovaltar: I get it! Looks better and shorter))  Thanks!

VitaliyBondarenko1982 (approved):
Looks good

PR #1846, opened 2024-02-22
adder done

anastasia1383:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
```
Rusik-best86, line 7: Try modifying the code to return the adder function once.

Rusik-best86 (changes requested):
Good job!
Please fix yout code.

src/makeInfinityAdder.js
anastasia1383: then it won't be a closure?

BudnikOleksii: > then it won't be a closure?

It remains the same, but it's unnecessary

BudnikOleksii (approved):
Good work 👍

PR #1848, opened 2024-02-23
solution

Dev1NA:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(value) {
```
l4st1m0za, line 11: The adder function only takes a single argument. It needs to be modified to accept multiple arguments.

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+    if (value === undefined) {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
+    sum += value;
```
l4st1m0za, line 19: The sum is calculated using a simple addition. It should be calculated using either 'forEach' or 'reduce' methods as per the task requirements.

l4st1m0za (changes requested):
Good job, but some changes are required

l4st1m0za (approved):
Good job!

PR #1854, opened 2024-02-26
Solution

olduma:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result = 0;
+  const adder = (...args) => {
```
Moroz-Dmytro, line 10: Why `...args` if you can have only 1 argument?

Moroz-Dmytro (changes requested):
Good work 👍 
But we can improve it

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+      result += number;
+
+      return adder;
+    } else {
+      const currentResult = result;
+
+      result = 0;
+
+      return currentResult;
+    }
```
BudnikOleksii, line 22:
```suggestion
    } 
    const currentResult = result;

    result = 0;

    return currentResult;
```
You don't need else as you return in the `if` block

BudnikOleksii (approved):
Good work 👍

PR #1855, opened 2024-02-26
Solution

ivan-gritsenko:


tetlisna (approved):
Looks good to me!

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,22 @@
...
+  let currentResult = 0;
+  let result = 0;
+
+  function adder(value) {
+    if (arguments.length !== 0) {
+      currentResult += value;
+
+      return adder;
+    }
+    result = currentResult;
```
BudnikOleksii, line 18:
```suggestion
  let currentResult = 0;

  function adder(value) {
    if (arguments.length !== 0) {
      currentResult += value;

      return adder;
    }
    const result = currentResult;
```
You use `result` only here so it can be `const`

BudnikOleksii (approved):
Good work 👍

PR #1856, opened 2024-02-26
add task solution

vladarskyi:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,30 @@
...
+
+      if (clearCall === 1) {
+        return result;
+      }
+
+      if (clearCall === 2) {
+        result = 0;
+
+        return result;
+      }
```
ivan-gritsenko, line 24:
```
if (clearCall !== 0) {
        const currentResult = result;

        result = 0;

        return currentResult;
      }
```

      You can avoid using secondary "if" construction by adding a temporary variable.

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,30 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let clearCall = 0;
+  let result = 0;
+
+  const adder = (arg) => {
+    if (arg === undefined) {
+      clearCall++;
+
+      if (clearCall === 1) {
```
ivan-gritsenko, line 16: try to avoid magic numbers

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,30 @@
...
+        return result;
+      }
+
+      if (clearCall === 2) {
```
ivan-gritsenko, line 20: same magic number

ivan-gritsenko (changes requested):
Good job! 
There are a few minor fixes left to make

src/makeInfinityAdder.js
vladarskyi: Well thank you. And in this way, I don't even need to count the number of function runs without arguments.

PR #1857, opened 2024-02-26
add solution

tetlisna:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,20 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (nextValue= null) => {
+    if (nextValue === null) {
```
BudnikOleksii, line 12:
```suggestion
  const adder = (nextValue) => {
    if (nextValue === undefined) {
```
It doesn't make much sense to assign null as default

BudnikOleksii (approved):
Good work 👍

PR #1860, opened 2024-02-29
solution

Anothar:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let b = 0;
+
+  const infinit = (a) => {
```
TarasHoliuk, line 11:
```suggestion
  const adder = (a) => {
```

src/makeInfinityAdder.js
```diff
@@ -6,7 +6,23 @@
...
+    if (a || a === 0) {
+      b = b + a;
+
+      return infinit;
+    }
+
+    const ba = b;
```
TarasHoliuk, line 18: it's just concatenated variable names? :) Let's replace a, b, ba with meaningful variable names
Even in this small function, it's easy to misunderstand something with such variables

TarasHoliuk (changes requested):

Esceype (approved):
Well done!

PR #1866, opened 2024-03-04
implemented the requirements

RostyslavKuzmych:


src/makeInfinityAdder.js
```diff
@@ -6,7 +6,27 @@
  */
 
 function makeInfinityAdder() {
-  // write code here
+  let result = 0;
+
+  const adder = (number) => {
+    return createFunction(number);
+  };
```
sarakhmen, line 13: I think you have some redundancies with inner functions here. Also pay attention to the redundant empty lines in your code. It doesn't increase readability.

sarakhmen (approved):
Good job, but you can improve your solution if you want ;)

PR #1876, opened 2024-03-11
solution

rwlp:


jv-aquino (changes requested):
Great job!! Concise solution and clean code! But please: add a more meaningful name to your variables!

jv-aquino (approved):
Good Rafael! 

And don't let yourself down for that, it's a good thing to be able to improve more and more!  In this case you just need more practice and examples of better naming so that it becomes automatic - and it will!

PR #1884, opened 2024-03-13
Solution

kravets111:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,27 @@
...
+  let sum = 0;
+  let currentSum = 0;
+
+  const ADDER = (num) => {
+    if (num !== undefined) {
+      currentSum += num;
+
+      return ADDER;
+    } else {
+      sum += currentSum;
```
etojeDenys, line 16:
```suggestion
    }
      sum += currentSum;
```
else is redundant here

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,27 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  let currentSum = 0;
+
+  const ADDER = (num) => {
```
etojeDenys, line 10: name functions in camelCase

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,27 @@
...
+    if (num !== undefined) {
+      currentSum += num;
+
+      return ADDER;
+    } else {
+      sum += currentSum;
+
+      const RESULT = sum;
```
etojeDenys, line 18: same about variables

etojeDenys (changes requested):

src/makeInfinityAdder.js
etojeDenys: image
still not fixed, feel free to ask me about it in the chat if you have some problems

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #1897, opened 2024-03-22
Solution

valost:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(num) {
+    if (num !== undefined) {
+      sum += num;
+
+      return adder;
+    } else {
```
etojeDenys, line 14: else is redundant here

etojeDenys (approved):
good job

PR #1906, opened 2024-04-02
adder

VerJane23:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,24 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+  const adder = (number) => {
```
GUSILLUS, line 8: add space

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,24 @@
...
+    }
+
+    if (number === undefined) {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
+  };
```
GUSILLUS, line 22: Let's write like that
```suggestion
    if (isNan(number) {
    	 const result = sum;

      sum = 0;

      return result;
    }
      sum += number;

      return adder;

  };
```

GUSILLUS (changes requested):
GJ! But:

src/makeInfinityAdder.js
VerJane23: ok, and add ')' for 'if'... however, now tests is broken

lerastarynets (approved):
lgtm 🚀

PR #1908, opened 2024-04-03
add task solution

AlexandeeerG:
Hellish challenge
When the word hell is present in the title of the task, nothing good should be expected. Let's get down to business. Write the makeAdder function, which will return the adder function.

The adder function should work on the following pattern: adder(2)(3)(4)(1)(2)(3)(4)() === 19. It will sum all the numbers passed in the parameters until it encounters a call without parameters.

When calling without parameters, it will return the result and clear the amount.

Example:

const adder = makeAdder();

console.log(
  adder(), // 0
  adder(4)(5)(), // 9
  adder(), // 0
  adder(5)(5)(5),
  adder(4),
  adder(), // 19
  adder(), // 0
);

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -124,9 +124,9 @@
       }
     },
     "@babel/eslint-parser": {
-      "version": "7.23.10",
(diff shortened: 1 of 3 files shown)


etojeDenys (approved):
good job

PR #1921, opened 2024-04-07
Solution

SlavaRevit:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,22 @@
...
-  // write code here
+  let RESULT = 0;
+  const adder = (...args) => {
+    if (args.length === 0) {
+      const value = RESULT;
+
+      RESULT = 0;
+
+      return value;
+    } else {
```
etojeDenys, line 15:
```suggestion
    }
```
else is redundant here

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,22 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let RESULT = 0;
```
etojeDenys, line 7:
```suggestion
  let result = 0;
```
it's not a constant, so better to name it in camelCase

etojeDenys (changes requested):
few comments

etojeDenys (approved):
lgtm

PR #1930, opened 2024-04-12
Solution

odunchyk:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(num) {
+    if (num !== undefined) {
+      sum += num;
+
+      return adder;
+    } else {
```
polosanya, line 14: `else` is redundant here as code will not go further after `return` anyway

polosanya (changes requested):
Almost done!

mrkovaliv (approved):

PR #1948, opened 2024-04-24
Solution

hardixxxxx:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
...
+      sum += number;
+
+      return infinityAdder;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
```
etojeDenys, line 20:
```suggestion
      return infinityAdder;
    }
    
      const result = sum;

      sum = 0;

      return result;
```
else is redundant here

etojeDenys (approved):
good job

PR #1984, opened 2024-05-23
Task js task infinite sum closures

HigorPedroso:


pedro-ruas (approved):
Well done! There is one minor thing to be attend of, but very good!

PR #1987, opened 2024-05-27
Solution

Pavlo-lab-max-Kopalko:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,25 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adder = (number) => {
+    if (number === undefined) {
```
vadiimvooo, line 10:
```suggestion
    if (!number) {
```

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,25 @@
...
+      sum = 0;
+
+      return result;
+    }
+
+    sum += number;
+
+    if (number !== undefined) {
+      return adder;
+    }
```
vadiimvooo, line 22: This line is redundant as you have one if statement above which handle this case

vadiimvooo (changes requested):
Good job. There are a few comments to fix below.

src/makeInfinityAdder.js
Pavlo-lab-max-Kopalko: hello, no. You unright in this moment. It should to left this condition. When function get 0 it understand him as a boolean, not like a number. What's why appeared a mistake. Now I need no add adding condition to do it works. I've left this condition as I do it at first.

nazarmatsevych (approved):
Good work!

PR #1989, opened 2024-05-31
 solution

MariCodec:


yaroslav1177 (approved):
well done!

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function addedFunc(a) {
```
roman-mirzoian, line 9: The name of the ``a`` parameter seems a bit generic.
It might be better to name it, for example, ``value`` to make it easier to understand.

roman-mirzoian (approved):
Well done and please pay attention to the comment.

PR #1992, opened 2024-06-02
made makeInfinityAdder

StsKushnir:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let currentSum = 0;
+
+  const adder = function (num) {
```
roman-mirzoian, line 9: This function can take any number of arguments, so using the ``num`` parameter limits it.
You can use arguments or the unpacking operator `(...args)`

Example:

```suggestion
  const adder = function (...args) {
```

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
...
+    if (num === undefined) {
```
roman-mirzoian, line 10: This num === undefined check doesn't work for a function that can take many arguments.
Here it is necessary to check whether the arguments are passed at all.

Example: 
```suggestion
    if (args.length === 0) {
```

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
...
+      const result = currentSum;
+
+      currentSum = 0;
+
+      return result;
+    } else {
+      currentSum += num;
```
roman-mirzoian, line 17: Adding numbers to currentSum is best done in a loop for all passed arguments.

Example:
```suggestion
      for (let num of args) {
        currentSum += num;
      }
```

roman-mirzoian (changes requested):
Good job, but pay attention to the comments.

roman-mirzoian (approved):
Well done!

PR #1994, opened 2024-06-09
Add task solution

Karanelus:
The `adder` function should work on the following pattern: `adder(2)(3)(4)(1)(2)(3)(4)() === 19`. It will sum all the numbers passed in the parameters until it encounters a call without parameters.


natalia-klonowska (approved):
Great work 🔥

PR #2002, opened 2024-06-22
Solution

siefimov:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let amount = 0;
+
+  const adder = (numberToAdd) => {
+    if (numberToAdd != null) {
```
VitaliyBondarenko1982, line 10: why comparing with null? and not strong comparing.
```suggestion
    if (numberToAdd !== undefined) {
```

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
...
+      amount += numberToAdd;
+
+      return adder;
+    } else {
```
VitaliyBondarenko1982, line 14: redundant else.
remove wrapping rest code in else.
in if block already exists return.

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some imrovements.

volodymyr-soltys97 (approved):
Well done! 🔥

PR #2009, opened 2024-07-09
solution

Kozubowicz:


Zibi95 (approved):
"It's working and looks good. It can even accept multiple parameters. Let's say you overdid it, but it's fine. Good job 🥇

PR #2011, opened 2024-07-10
Solution

objelov:


src/makeInfinityAdder.js
```diff
@@ -5,6 +5,23 @@
  */
 function makeInfinityAdder() {
   // write code here
```
4ksandrey, line 7: remove this comment)

src/makeInfinityAdder.js
```diff
@@ -5,6 +5,23 @@
...
+  let result = 0;
+
+  const adder = (...args) => {
```
4ksandrey, line 10: think about naming this variable much better and understandable

4ksandrey (changes requested):
Check the comments and fix minor mistakes

src/makeInfinityAdder.js
objelov: fixed

src/makeInfinityAdder.js
objelov: Fixed)

As I understood it is about `adder` variable, right?

chernobrovin (approved):

PR #2013, opened 2024-07-13
solution

OPTIMISTIXX:


volodymyr-soltys97 (approved):
Good job!
For the future you need to create a new branch for your solution 
image

PR #2024, opened 2024-07-17
Solution

AndriiZakharenko:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,22 @@
...
+      sum = sum + param;
+    } else {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    }
+
+    return adder;
```
Moroz-Dmytro, line 19:
```suggestion
    if (param !== undefined) {
      sum += param;
      
      return adder;
    }
    const result = sum;

    sum = 0;

    return result;
```

Moroz-Dmytro: We can avoid using `if/else`

Moroz-Dmytro (approved):

src/makeInfinityAdder.js
AndriiZakharenko: done

PR #2029, opened 2024-07-19
add solution

RimmaBuzuluk:


AndriiZakharenko (approved):
good job!

piyopiyo0 (approved):
Looks great! 👍

src/makeInfinityAdder.js
```diff
@@ -1,10 +1,24 @@
 'use strict';
 
-/***
- * @return {function}
- */
+// /***
+//  * @return {function}
+//  */
+
```
VenomMB, line 6: remove it

src/makeInfinityAdder.js
```diff
@@ -1,10 +1,24 @@
...
 function makeInfinityAdder() {
-  // write code here
+  let count = 0;
+
+  return function countFunction(a) {
```
VenomMB, line 10:
```suggestion
  return function countFunction(value) {
```

src/makeInfinityAdder.js
```diff
@@ -1,10 +1,24 @@
...
+    if (a === undefined) {
```
VenomMB, line 11:
```suggestion
    if (!value) {
```

src/makeInfinityAdder.js
VenomMB, line 10:
```suggestion
  return function countCall(a) {
```

VenomMB (changes requested):
Great job, leave small comments

src/makeInfinityAdder.js
RimmaBuzuluk: hello, I can not do it because, sometimes value = 0.
image

BudnikOleksii (approved):
👍

PR #2068, opened 2024-08-26
Solution

przwojwwp:


src/makeInfinityAdder.js
```diff
@@ -5,6 +5,23 @@
...
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
+      sum += x;
+
+      return adder;
+    }
```
natalia-klonowska, line 21: you don't need `else` because you have `return` in `if` statement

natalia-klonowska (changes requested):

natalia-klonowska (approved):

PR #2073, opened 2024-08-30
solved

Artem-Brui:


Artem-Brui:
change isNaN to underfind

anastasiiavorobiova (approved):
Great work!

PR #2082, opened 2024-09-06
add task solution

ipaiareli:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,21 @@
...
+  let total = 0;
+
+  return function adder(value) {
+    if (value === undefined) {
+      const result = total;
+
+      total = 0;
+
+      return result;
+    } else {
```
etojeDenys, line 16:
```suggestion
    }
```
else is redundant here

etojeDenys (approved):
good job

PR #2114, opened 2024-10-11
solution

DemianZolotukhin:


Pirgei (changes requested):
Please fix the comments left by the AI mentor. Remove the variable copyOfSum, as it can be declared directly within the if condition.
Remove the unnecessary else, because after the return statement, the function stops execution, making the code that follows unreachable

PR #2142, opened 2024-11-09
resolve task

Daniil-102:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,21 @@
  * @return {function}
  */
 function makeInfinityAdder() {
```
OkMoroz, line 6: Згідно файлу readme.md _"Let's get down to business. Write the `**makeAdder**` function, which will return the `adder` function."_

OkMoroz (approved):
**Good job!** 👏 Please take a look at my comment, thank you!

BudnikOleksii (approved):
Good job 👍

PR #2144, opened 2024-11-09
Solution

OkMoroz:


Krykunov (approved):
Good job! 👍

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
...
+  let sum = 0;
+
+  function adder(arg) {
+    if (arg === undefined) {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
```
BudnikOleksii, line 16: You don't need `else` as you return in `if` statement

BudnikOleksii (approved):
Good job 👍

PR #2151, opened 2024-11-10
solution

minimal005:


chelsea7smile (approved):
I like your solution, especially name of function `adder` (I used just `add` ;) ) ;
Good work 🚀

src/makeInfinityAdder.js
```diff
@@ -1,10 +1,24 @@
...
+  let sum = 0;
+
+  return function adder(number) {
+    if (number === undefined) {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
```
BudnikOleksii: You don't need `else` as you return in `if` statement

BudnikOleksii (approved):
Good job 👍

PR #2153, opened 2024-11-11
solution

k0tlik02:
Hellish challenge

### DIFF
--- a/src/makeInfinityAdder.js
+++ b/src/makeInfinityAdder.js
@@ -4,7 +4,21 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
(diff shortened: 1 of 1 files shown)


taniabarkovskya (approved):
good job! you can be proud of yourself ⭐

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,21 @@
...
+  let sum = 0;
+
+  return function adder(number) {
+    if (number === undefined) {
+      const result = sum;
+
+      sum = 0;
+
+      return result;
+    } else {
```
BudnikOleksii, line 16: You don't need `else` as you return in `if` statement

BudnikOleksii (approved):
Good job 👍

PR #2154, opened 2024-11-11
add task solution

taniabarkovskya:


Krykunov (approved):
Good job! 👍

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,21 @@
...
+      const currentSum = sum;
+
+      sum = 0;
+
+      return currentSum;
+    } else {
+      sum += a;
+
+      return add;
+    }
```
BudnikOleksii, line 20: You don't need `else` as in `if` you return

BudnikOleksii (approved):
Good work 👍

PR #2157, opened 2024-11-11
Solution

iirk1:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,29 @@
...
+      const res = sum;
+
+      sum = 0;
+
+      return res;
+    }
+
+    if (argument === 0) {
+      return adder;
+    }
```
TarasHoliuk:
```suggestion
```
I agree - you can remove it:

image

TarasHoliuk (changes requested):
One small change is required

VitaliyBondarenko1982 (approved):
Looks good

PR #2270, opened 2025-02-21
chainer solution

reb0rned:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function chainer(el) {
```
igorkruz, line 9: Choose a more descriptive name instead of `el`

igorkruz (approved):
Great job 👍 
Please consider the added comments.

PR #2347, opened 2025-05-02
add task infinite sum closures solution

DenisPitsul:


OksanaKuziv (approved):
this looks like a good job)

PR #2348, opened 2025-05-03
task solution

NikD0R:


NikD0R:
@mateacademy-ai-mentor 
This if (typeof elem !== 'number') need to check if the argument wasn't given

PR #2354, opened 2025-05-06
solution added

RTSaitama:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,22 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let counter = 0;
```
hryhoriikorsun, line 7: It's more about sum not counting. 
You can name this variable like `sum` but also `counter` not bad too.

hryhoriikorsun (approved):
Great job💪

hryhoriikorsun (approved):

PR #2390, opened 2025-06-17
res

O1ympus:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,25 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let res = 0;
```
ArturErmolenko, line 7:
```suggestion
  let result = 0;
```

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,25 @@
...
+  function adder(...params) {
+    if (params.length === 0) {
+      const result = res;
+
+      res = 0;
+
+      return result;
+    }
+
+    for (const el of params) {
```
ArturErmolenko, line 18:
```suggestion
    for (const param of params) {
```

ArturErmolenko (approved):
Don't be lazy ;)
Good naming is required

PR #2414, opened 2025-07-16
Solution

MaxOnRem:
function sums all the numbers passed in the parameters until it encounters a call without parameters. Using chaining and closures


anastasiiavorobiova (approved):
⭐⭐⭐⭐⭐ 
Super!

PR #2417, opened 2025-07-19
solved

kshyndina:


kshyndina:
@mateacademy-ai-mentor omg, it was like that in the task after i cloned it. fix your tasks before asking for such corrections from students.

PR #2419, opened 2025-07-21
Task solution

horidor:
Added task solution.


O1ympus (changes requested):
use arguments.length===0 check instead of a === undefined

O1ympus (approved):

PR #2422, opened 2025-07-22
solution for infinite sum closures task

Aonarro:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  const adderFunction = (...args) => {
```
YevheniiKa, line 9: As far as I know, ...args iterates through all elements in the array, but our input parameter is a number.

src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
...
+    if (args.length === 0) {
```
YevheniiKa, line 10: I would also try adding a check for typeof === ‘number’ and for falsy cases, such as null, undefined, and others, instead of length.

YevheniiKa (approved):
Great Job!

PR #2477, opened 2025-09-19
Solution

Julia-Hertseva:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,23 @@
...
+  let sum = 0;
+
+  function adder(arg) {
+    if (typeof arg === 'undefined') {
+      const printed = sum;
+
+      sum = 0;
+
+      return printed;
+    } else {
```
DmytroLymar, line 16: You can just write
`sum += arg;
 return adder;`
 
 without `else` or check for arg is number

DmytroLymar (approved):
Good job, simple and clear. Just one comment, but it's not critical

PR #2480, opened 2025-09-21
Solution

DmytroLymar:


src/makeInfinityAdder.js
```diff
@@ -4,7 +4,25 @@
  * @return {function}
  */
 function makeInfinityAdder() {
-  // write code here
+  let sum = 0;
+
+  function adder(num) {
+    if (arguments.length === 0) {
```
Julia-Hertseva, line 10: The use of arguments.length === 0 is valid, but you might use typeof num === 'undefined' for clarity and proofing

Julia-Hertseva (approved):
Great use of  closure! 💪

PR #2502, opened 2025-10-14
Solution

pzwolenik:


pzwolenik:
@mateacademy-ai-mentor all test are for function with a name makeInfinityAdder. I won't implemented your suggestions to rename it to makeAdder.

pzwolenik:
@mateacademy-ai-mentor test required name makeInfinityAdder. I'm sure I have the right solution.

PR #2543, opened 2025-12-21
solution

artemstadnik:


dmytro-halieba (approved):
Great job! Overall, the code is functionally correct according to the task and is clear to follow.

Optional improvements include adding type checking: currently, no type checking is performed on the input number; passing non-numeric values may produce unexpected results. This can be improved by verifying `typeof number === 'number'`.

PR #2610, opened 2026-04-02
Solution v01

Kostiantyn-Sacheiev:


Kostiantyn-Sacheiev:
@mateacademy-ai-mentor
The main function names makeInfinityAdder(). Also I have the module.exports = makeInfinityAdder; in the source file
I named the inner function a makeAdder. But the name of the main function is the totaly correct

